PR #94, opened 2019-04-15
Image Service - Single Image

mikeapp:
Recipe #5

### DIFF
--- a/cookbook/0005-image-service/index.md
+++ b/cookbook/0005-image-service/index.md
@@ -16,7 +16,9 @@ Annotate a Canvas using an image with an associated IIIF Image API service.
 
 ## Implementation notes
 
-The annotation structure follows that of the [Simplest Manifest - Image] recipe. Within the `body` of the image annotation, specify the IIIF Image API service using the `service` property. The `id` of the service is the base URI of the associated IIIF Image API service.    
+The annotation structure follows that of the [Simplest Manifest - Image] recipe. Within the `body` of the image annotation, specify the IIIF Image API service using the `service` property. The `id` of the service is the base URI of the associated IIIF Image API service.  
+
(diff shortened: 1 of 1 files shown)


mikeapp:
#5 lists two 2.1 fixtures - one with a simple image service, the other contains additional properties for the service.  I'm unsure how we want to handle this - split the issue or have two examples in recipe, with additional explanation.

cookbook/0005-image-service/index.md
```diff
@@ -0,0 +1,83 @@
...
+  ],
+  "id": "https://example.org/iiif/book5/manifest",
+  "type": "Manifest",
+  "label": { "en": [ "Simple Manifest - Image with IIIF Image API Service" ] },
+  "items": [
+    {
+      "id": "https://example.org/iiif/book5/canvas/p1",
+      "type": "Canvas",
+      "height": 1800,
+      "width": 1200,
```
stephenwf: Should we add a label to the canvas (marked as recommended in the specification)

mikeapp:
Probably.  I copied the example from mvm-image, which also lacks a label - but "minimum viable" is not equivalent to best practice.

cookbook/0005-image-service/index.md
mikeapp: Probably. I copied the example from mvm-image, which also lacks a label - but "minimum viable" is not equivalent to best practice.

stephenwf: That makes sense 👍

mikeapp:
Per discussion on IIIF/api#1769 expand on annotation `id` needing to reference an image but not necessarily derived from a Image API service.

mikeapp:
Also duplicated by @tomcrane  in themed branch

PR #95, opened 2019-04-18
The most minimal Jekyll, with prism.js

tomcrane:


tomcrane:
For this to work, we'll need to enable GitHub pages for the master branch, but I think that's OK.

tomcrane:
Yes, we will - but this allows you to see line numbers, highlighted lines etc while you're working on a recipe. I'll add a comment to https://github.com/IIIF/api/issues/1758 with more detail.

Otherwise you can't really see what it looks like until built.

PR #96, opened 2019-04-19
Use recipe frontmatter to automate parts of cookbook

tomcrane:
A hefty reorg to get the paths correct for later integration with the website, and also use recipe frontmatter where possible.


tomcrane:
I'm going to merge this as the current master is broken anyway, for path previews

PR #98, opened 2019-05-03
Adding deployment to AWS

glenrobson:
Currently deploys to:

https://preview.iiif.io/cookbook/<branch_name>/

see `https://preview.iiif.io/cookbook/deployment/` as an example. Once this is merged all non-master branches will deploy to preview. lll have to do some more thinking around the master branch.

@tomcrane I renamed README.md to index.md to make it work when you opened the `/cookbook/<branch_name>/`. I then created a `README.md` which is more to do with deployment but in the long run I think would be a good place to document the process for submitting recipes to github. Are you happy with that change?


tomcrane:
@glenrobson I think we could get these merged even if we want to tweak the process some more later

glenrobson:
Thanks both

Issue #103, opened 2019-09-17
Annotating a Poetry Reading


saracarl:
## Annotating a Poetry Reading

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
While an audio file of a poetry performance may be divided into a track for each poem, scholars may wish to use annotations to indicate aspects of the performance of a particular poem.  

Examples of the kinds of information to be annotated may include the following:
* structural information (introduction, title, stanzas)
* points where the performer takes a breath
* repeated phrases

saracarl:
We should add this to 
https://github.com/BrumfieldLabs/cookbook-recipes/blob/master/recipe/index.md#other-kinds-of-annotations
when we're done.

saracarl:
To Test:  If we have web annotations in a separate file, can we just include them in the recipe's directory?  Will they get built?   How and where do we references them (outside of the manifest?)

glenrobson:
Yes the annotations file should be in the same directory as the manifest. Regarding referencing them please give me a shout when this is ready so we can try a few options.

glenrobson:
Done in: https://github.com/IIIF/cookbook-recipes/pull/130

PR #107, opened 2019-10-28
Recipe #103 -- Annotating a Poetry Reading

saracarl:
An initial attempt at a IIIF v3 manifest with annotations on an audio file.

This needs to be updated with github pages references instead of raw github references to our repo, but I think that's a chicken & egg problem that will be fixed after merging (by whom?)

Discussion here or in the IIIF AV Slack channel.

### DIFF
--- a/recipe/0103-poetry-reading-annotations/index.md
+++ b/recipe/0103-poetry-reading-annotations/index.md
@@ -29,7 +29,7 @@ This recipe shows two variations of referencing annotations from within a manife
 
 ## Example
 
-A manifest for the poem "Her Kind" read by Anne Sexton in 1974.  The recording is 107 minutes long.  Annotations are included in the manifest and show both a point annotation (at a particular time) and a range annotation (covering a time range).  (For more on these annotations, see the [annotation.json file description.](#annotations))
+A manifest for the poem "Her Kind" read by Anne Sexton in 1974.  The recording is 107 seconds long.  Annotations are included in the manifest and show both a point annotation (at a particular time) and a range annotation (covering a time range).  (For more on these annotations, see the [annotation.json file description.](#annotations))
 
 {: .line-numbers data-src="manifest1.json" }
 ```json
--- a/recipe/0103-poetry-reading-annotations/index.md
+++ b/recipe/0103-poetry-reading-annotations/index.md
@@ -23,9 +23,14 @@ This implementation builds off of the [audio example][0002], but adds Web Annota
 
 ## Example
 
-Describe in prose and provide examples, e.g.: 
+A manifest for the poem "Her Kind" read by Anne Sexton in 1974.  The recording is 107 minutes long.
 
 {: .line-numbers data-src="manifest.json" }
+```json
+```
+Annotation examples, both a point annotation (at a particular time) and a range annotation (covering a time range)
+
+{: .line-numbers data-src="annotations.json" }
 ```json
 ```
 
--- a/recipe/0103-poetry-reading-annotations/index.md
+++ b/recipe/0103-poetry-reading-annotations/index.md
@@ -1,0 +1,40 @@
+---
+title: Annotating a Poetry Reading
+id: 103
+layout: recipe
+tags: [audio, presentation, annotation]
+summary: "Use annotations to indicate aspects of the performance of a particular poem."
+---
+
+
+## Use Case
+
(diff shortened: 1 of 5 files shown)


glenrobson:
Manifest validates OK with new 3.0 validator: https://iiif.io/api/presentation/validator/service/validate?url=https://raw.githubusercontent.com/IIIF/cookbook-recipes/1ab053f8abb678488b86b9b1ecf98681f35b58c4/recipe/0103-poetry-reading-annotations/manifest1.json&version=3.0 

(Probably more of a test for my validator code :-)).

glenrobson:
Manually deploy to: https://preview.iiif.io/cookbook/Recipe_103_brumfieldlabs/recipe/0103-poetry-reading-annotations/

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,58 @@
...
+## Implementation notes
+
+This implementation builds off of the [audio example][0002], but adds Web Annotations.
+
+This recipe shows two variations of referencing annotations from within a manifest.  The first has the annotations embedded within the manifest file.  The second has a reference to annotations in a separate file.
+
+
+## Example
+
+A manifest for the poem "Her Kind" read by Anne Sexton in 1974.  The recording is 107 minutes long.  Annotations are included in the manifest and show both a point annotation (at a particular time) and a range annotation (covering a time range).  (For more on these annotations, see the [annotation.json file description.](#annotations))
```
cjcolvar:
```suggestion
A manifest for the poem "Her Kind" read by Anne Sexton in 1974.  The recording is 107 seconds long.  Annotations are included in the manifest and show both a point annotation (at a particular time) and a range annotation (covering a time range).  (For more on these annotations, see the [annotation.json file description.](#annotations))
```
I think you meant to put seconds here instead of minutes.

azaroth42:
The annotations should be part of the Canvas, not the Manifest.

saracarl:
Add frontmatter for tags and summary, similar to PR #110 : 

`tags: [video, presentation, opera]`
`summary: "Complex nested table of contents for an opera."`

saracarl:
Replace URLs with magic URLs for IDs.

saracarl:
annotation motivation should be an array.

saracarl:
Where we use "RangeSelector" for the annotation target, we could instead use Media Fragments "#t=1.23,2.23" appended to the canvas URL.  Instead of "source" you'd condense it into the "target" field.
We should just document this as an alternative, but not change our example.

saracarl:
Ok, I believe this is ready for a final review and sponsorship to the TRC.  
Changes:
1)  Annotations are part of the canvas, not manifest.
2)  Frontmatter with a summary & tags
3)  Magic URL replacement.  This needs another build and check because in order to get them looking right I had to use:
`{{site.url}}{{site.baseurl}}/{{page.path}}`
instead of
`{{site.url}}\{{site.baseurl}}/{{page.path}}`
(note the removal of the \ between site.url and site.baseurl)
4)  Implementation notes on arrays for annotation motivation and range selector vs media fragments.

@glenrobson Would you look at #3 above and make sure this gets "processed" appropriately -- organizationally and technically?   Thanks!

azaroth42:
Is there a reason that the Canvases have URIs that end in `canvas/segment`? While they are just opaque URIs, it gives the impression that they're parts of canvases, not the whole thing.

I think it would be good to remove all the `segment` endings from the canvas URIs.

saracarl:
I think that is a mistake in my search/replace for the magic URLs.  I'll fix it.

glenrobson:
Moving this work to https://github.com/IIIF/cookbook-recipes/pull/130 so closing this pull request. Please comment on 130.

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+Both are correct.
+
+4. While the IIIF Specification requires arrays for most values, the W3C Web Annotation Specification does not require arrays, so the following line: 
+
+`"motivation": "commenting",`
+
+is correct, but unusual for a IIIF resource.  It could also be expressed (and must be if there is more than one motivation) as follows:
+
+`"motivation": ["commenting"],`
```
benwbrum: If the IIIF spec requires arrays for motivations, should we remove this implementation note altogether?

PR #123, opened 2019-11-06
Update recipe 2

azaroth42:
Mostly in the example, some tweaks to the recipe content.


glenrobson:
(Automated comment): Branch aviliable for testing at https://preview.iiif.io/cookbook/0002-simplest-audio-update

zimeon:
Specific page on preview: https://preview.iiif.io/cookbook/0002-simplest-audio-update/recipe/0002-mvm-audio/

glenrobson:
I've updated the Ids so they use the new Jekyll variables and also linked to a real audio file hosted in the iiif s3 bucket. 

Does this need to go through TRC or can it be approved as its one of the initial cases?

tomcrane:
This example was essentially published as an example to get the cookbook started. I don't think it needs to go to TRC, but don't object if it does.

tomcrane:
I also think it should be merged - it's already in `master`, this PR does not change the text of the recipe - it provides some essential fixes.

tomcrane:
I made one further tweak in the above commit.

glenrobson:
(Automated comment): Branch available for testing at https://preview.iiif.io/cookbook/0002-simplest-audio-update

glenrobson:
Added a link to the UV but get the following error when opening:

"Unable to determine media type"

glenrobson:
Looks like the audio support is an issue with the UV rather than the recipe: https://github.com/UniversalViewer/universalviewer/issues/730

Merging this as discussed in the editors channel as existing live recipe doesn't point to a working audio file.

PR #128, opened 2019-11-20
Clean jekyll magic

glenrobson:
Creating a plugin so we can have clear Jekyll identifiers and don't need front matter. Implemented the following jekyll variables:

Full URL to current file:
```
"id": "{{ id.url }}"
```

Full URL to current directory:
{{ id.basename }}

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ cache: bundler
 before_script:
   - bundle install
 script:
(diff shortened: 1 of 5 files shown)


glenrobson:
(Automated comment): Branch aviliable for testing at https://preview.iiif.io/cookbook/clean_jekyll_magic

PR #130, opened 2019-11-20
Recipe #103 -- Annotating a Poetry Reading

glenrobson:
Merging Sara's remote copy to the IIIF repo so that it can be built automatically. Original pull request with comments is https://github.com/IIIF/cookbook-recipes/pull/107. Original description from Sara:
```
An initial attempt at a IIIF v3 manifest with annotations on an audio file.

This needs to be updated with github pages references instead of raw github references to our repo, but I think that's a chicken & egg problem that will be fixed after merging (by whom?)

Discussion here or in the IIIF AV Slack channel.
```


glenrobson:
(Automated comment): Branch aviliable for testing at https://preview.iiif.io/cookbook/0103-poetry

glenrobson:
Actually recipe available at:

https://preview.iiif.io/cookbook/0103-poetry/recipe/0103-poetry-reading-annotations/

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,96 @@
...
+              "body":{
+                "id":"https://library.harvard.edu/poetry/audio/listeningbooth//PS3537E915A6x1974/Her_Kind.mp3",
+                "type":"Sound",
+                "format":"audio/mp3",
+                "duration":107
+              },
+              "target":"{{ id.path }}/canvas/1"
+            }
+          ],
+          "annotations":[
```
azaroth42: This is not correct. `annotations` is not allowed on an AnnotationPage, per the spec.

recipe/0103-poetry-reading-annotations/annotations.json
```diff
@@ -0,0 +1,44 @@
+{
+  "@context": "http://iiif.io/api/presentation/3/context.json",
+  "id": "{{ id.url }}",
+  "type": "AnnotationPage",
+  "items": 
+  [
+   { 
+      "@context":"http://www.w3.org/ns/anno.jsonld",
```
azaroth42: This `@context` should not be here.

recipe/0103-poetry-reading-annotations/annotations.json
```diff
@@ -0,0 +1,44 @@
...
+      "target":{ 
+         "source":"{{ id.path }}/canvas/1",
+         "selector":{ 
+            "type":"PointSelector",
+            "t":"27.660653"
+         }
+      }
+   },
+  {
+    "@context": "http://www.w3.org/ns/anno.jsonld",
```
azaroth42: Nor this.

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,96 @@
...
+            {
+              "id":"{{ id.path }}/annotations.json",
+              "type":"AnnotationPage",
+              "items":[
+                {
+                  "@context":"http://www.w3.org/ns/anno.jsonld",
```
azaroth42: Nor this

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,96 @@
...
+                  "target":{
+                    "source":"{{ id.path }}/canvas/1",
+                    "selector":{
+                      "type":"PointSelector",
+                      "t":"27.660653"
+                    }
+                  }
+                },
+                {
+                  "@context":"http://www.w3.org/ns/anno.jsonld",
```
azaroth42: And nor this.

recipe/0103-poetry-reading-annotations/manifest2.json
```diff
@@ -0,0 +1,48 @@
...
+          {
+            "id": "{{ id.path }}/canvas/1/paintings",
+            "type": "AnnotationPage",
+            "items": [
+              {
+                "id": "{{ id.path }}/canvas/1/painting/1",
+                "type": "Annotation",
+                "motivation": "painting",
+                "body": {
+                  "id": "https://library.harvard.edu/poetry/audio/listeningbooth//PS3537E915A6x1974/Her_Kind.mp3",
```
azaroth42: // in URL

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,96 @@
...
+                  "body":{
+                    "type":"TextualBody",
+                    "value":"breath",
+                    "format":"text/plain"
+                  },
+                  "target":{
+                    "source":"{{ id.path }}/canvas/1",
+                    "selector":{
+                      "type":"PointSelector",
+                      "t":"27.660653"
```
glenrobson: According to this: https://iiif.io/api/annex/openannotation/#point-selector t should be a floating point number.

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,96 @@
...
+                  "body":{
+                    "type":"TextualBody",
+                    "value":"her kind",
+                    "format":"text/plain"
+                  },
+                  "target":{
+                    "source":"{{ id.path }}/canvas/1",
+                    "selector":{
+                      "type":"RangeSelector",
+                      "t":"46.734653,47.875068"
```
glenrobson: According to this: https://www.w3.org/TR/annotation-model/#range-selector

I think this should be:

```
"selector": {
      "type": "RangeSelector",
      "startSelector": {
        "type": "PointSelector",
        "t":46.734653
      },
      "endSelector": {
        "type": "PointSelector",
        "t": 47.875068
      }
    }
```

recipe/0103-poetry-reading-annotations/manifest2.json
glenrobson: This is now fixed with Sara's commit: https://github.com/IIIF/cookbook-recipes/pull/130/commits/e3e42c7bb422c65595e2fcfd5ca51a04f873cf98

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: This is now resolved with: https://github.com/IIIF/cookbook-recipes/pull/130/commits/107ecabf6c30a7970bb03ebc42ccd12cb9fc1816

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: This is now resolved with: https://github.com/IIIF/cookbook-recipes/pull/130/commits/107ecabf6c30a7970bb03ebc42ccd12cb9fc1816

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: This is now resolved by https://github.com/IIIF/cookbook-recipes/pull/130/commits/e3e42c7bb422c65595e2fcfd5ca51a04f873cf98

recipe/0103-poetry-reading-annotations/annotations.json
```diff
@@ -0,0 +1,42 @@
...
+    "body": {
+      "type": "TextualBody",
+      "value": "her kind",
+      "format": "text/plain"
+    },
+    "target": {
+      "source": "{{ id.path }}/canvas/1",
+      "selector": {
+        "type": "RangeSelector",
+        "t": "46.734653,47.875068"
```
glenrobson: According to this: https://www.w3.org/TR/annotation-model/#range-selector

I think this should be:

"selector": {
      "type": "RangeSelector",
      "startSelector": {
        "type": "PointSelector",
        "t":46.734653
      },
      "endSelector": {
        "type": "PointSelector",
        "t": 47.875068
      }
    }

recipe/0103-poetry-reading-annotations/manifest1.json
saracarl: tab problem.  annotations is a peer of annotationpage.  Fixed.

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,84 @@
...
+2. This recipe shows two variations of referencing annotations from within a manifest.  The first has the annotations embedded within the manifest file.  The second has a reference to annotations in a separate file.
+
+3. Where we use "RangeSelector" for the annotation target, we could instead use a Media Fragment like "#t=1.23,2.23" appended to the canvas URL. Instead of "source" you'd condense it into the "target" field.  For example:
+
+```
+                "target": {
+                  "source": "http://localhost:4000/recipe/0103-poetry-reading-annotations/manifest1.json/canvas/segment1/canvas/segment1",
+                  "selector": {
+                    "type": "RangeSelector",
+                    "t": "46.734653,47.875068"
```
glenrobson: Sorry Sara, this also needs to be changed to:

```
"selector":{
                                 "type":"RangeSelector",
                                 "startSelector":{
                                    "type":"PointSelector",
                                    "t":46.734653
                                 },
                                 "endSelector":{
                                    "type":"PointSelector",
                                    "t":47.875068
                                 }
                              }
```

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,101 @@
...
+                     "body":{
+                        "id":"https://library.harvard.edu/poetry/audio/listeningbooth//PS3537E915A6x1974/Her_Kind.mp3",
+                        "type":"Sound",
+                        "format":"audio/mp3",
+                        "duration":107
+                     },
+                     "target":"{{ id.path }}/canvas/1"
+                  }
+               ],
+               "annotations":[
```
glenrobson: I think this still needs to be part of the canvas rather than the AnnotationPage. In practice this means adding:
```
      }
],
```
after line 32 and removing the extra `}` and `]` on line 82 and 83.

recipe/0103-poetry-reading-annotations/manifest2.json
```diff
@@ -0,0 +1,58 @@
...
+                        "duration":107
+                     },
+                     "target":"{{ id.path }}/canvas/1"
+                  }
+               ]
+            }
+         ]
+      }
+   ],
+   "annotations":[
```
glenrobson: This 'annotations' field also I think should be part of the canvas not part of the manifest...

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: I believe this is now fixed. The only `@context` in manifest1.json is at the top.

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: As above now fixed.

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: Now fixed

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: Now fixed.

recipe/0103-poetry-reading-annotations/index.md
glenrobson: Now fixed

recipe/0103-poetry-reading-annotations/manifest2.json
glenrobson: Now fixed.

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: Now fixed

azaroth42:
Comments on the text:
* The numbered list is broken -- it restarts at 1. for "While the IIIF Specification" ...
* The examples do not use arrays of motivations, so I suggest deleting the entry completely.
* The text and example do not line up -- The text says the final example has both supplementing and commenting, whereas it only has commenting.  As below, I think it should be only supplementing.

Comments on the examples:
* Shouldn't the MP3 be stored in IIIF S3 space to ensure that it stays working (see Newspaper recipe)
* Is "her kind" a transcription of the spoken content? If so then it should be motivation supplementing, not commenting. See: https://iiif.io/api/presentation/3.0/#35-values

glenrobson:
Thanks for your comments Rob. Regarding:

> Shouldn't the MP3 be stored in IIIF S3 space to ensure that it stays working (see Newspaper recipe)

We did discuss this but haven't heard from Harvard if they are happy for us to take a copy of the content so haven't done so. It would be interesting to raise in the TRC what peoples thoughts are on this but is it better to have examples with external content soon or examples with hosted content which may take longer to arrange? 

For info there was a similar discussion with the Newspapers and thankfully Europeana were OK for us to copy the images but we only copied the first two images and left the last two as remote.

recipe/0103-poetry-reading-annotations/annotations.json
```diff
@@ -0,0 +1,48 @@
+{
+   "@context":"http://iiif.io/api/presentation/3/context.json",
+   "id":"{{ id.url }}",
+   "type":"AnnotationPage",
+   "items":[
+      {
+         "id":"{{ id.path }}/canvas/annotation1",
+         "type":"Annotation",
+         "motivation":"commenting",
```
glenrobson: Per Rob's comment "Is "her kind" a transcription of the spoken content? If so then it should be motivation supplementing, not commenting. See: https://iiif.io/api/presentation/3.0/#35-values"

This should be:
```
"motivation":["supplementing"]
```

recipe/0103-poetry-reading-annotations/annotations.json
```diff
@@ -0,0 +1,48 @@
...
+            "selector":{
+               "type":"PointSelector",
+               "t":27.660653
+            }
+         }
+      },
+      {
+         "id":"{{ id.path }}/canvas/segment1/annotation/4",
+         "type":"Annotation",
+         "motivation":"commenting",
```
glenrobson: Per Rob's comment "Is "her kind" a transcription of the spoken content? If so then it should be motivation supplementing, not commenting. See: https://iiif.io/api/presentation/3.0/#35-values"

Listening to the recording it is a transcription so this should be:
```
"motivation":["supplementing"]
```

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,101 @@
...
+      ],
+      "annotations":[
+        {
+          "id":"{{ id.path }}/annotations.json",
+          "type":"AnnotationPage",
+          "items":[
+            {
+              "id":"{{ id.path }}/canvas/1/annotation/1",
+              "type":"Annotation",
+              "motivation":"commenting",
```
glenrobson: Should this be supplementing as above?

recipe/0103-poetry-reading-annotations/manifest1.json
```diff
@@ -0,0 +1,101 @@
...
+                "selector":{
+                  "type":"PointSelector",
+                  "t":27.660653
+                }
+              }
+            },
+            {
+              "id":"{{ id.path }}/canvas/1/annotation/2",
+              "type":"Annotation",
+              "motivation":"commenting",
```
glenrobson: Per Rob's comment "Is "her kind" a transcription of the spoken content? If so then it should be motivation supplementing, not commenting. See: https://iiif.io/api/presentation/3.0/#35-values"

Listening to the recording it is a transcription so this should be:
```
"motivation":["supplementing"]
```

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,91 @@
...
+```json
+```
+
+The same manifest, but with a reference to annotations in a separate file.  The annotation file is included below the manifest. 
+
+{: .line-numbers data-src="manifest2.json" }
+```json
+```
+
+<a name="annotations"></a>Annotation examples, both a point annotation (at a particular time) and a range annotation (covering a time range).  We use the "supplementing" motivation because these annotations are derived from the audio file on the canvas and the "commenting" motivation because the annotations are a comment about the canvas.  Annotations must have "supplementing" as a motivation for any annotations derived from the items on the canvas.
```
glenrobson: Related to the other changes but Rob's comments:

"The text and example do not line up -- The text says the final example has both supplementing and commenting, whereas it only has commenting. As below, I think it should be only supplementing."

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: but should be an array of values.

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: but needs to be an array

recipe/0103-poetry-reading-annotations/manifest1.json
saracarl: We're going to change "her kind" into "repetition of 'her kind'" -- it's meant as commentary on the structure of the poem, rather than a direct transcription.  That will clarify it.

recipe/0103-poetry-reading-annotations/annotations.json
saracarl: We'll change this to an array so it matches the IIIF spec rather than the w3c annotation spec.

Commenting is the correct motivation, as below.

recipe/0103-poetry-reading-annotations/annotations.json
saracarl: Comment.  Pretty sure our academic partner calls this "paratext", which I wouldn't call transcription.

benwbrum:
I'm trying to figure out whether the motivations in this example should be `["supplementing", "commenting"]` or just `["commenting"]`.  @saracarl has cleared up the fact that the annotations are not supposed to be transcriptions, which would necessitate a `supplementing` motivation as I understand it, but the annotations are certainly derived from the canvas, so I'm not sure whether `supplementing` should be added.

The [IIIF spec](https://iiif.io/api/presentation/3.0/#53-canvas) says, 

> Thus content of any type may be associated with the Canvas via an Annotation that has the `motivation` value `painting`, meaning the content is part of the Canvas; an Annotation that has the `motivation` value `supplementing`, meaning the content is from the Canvas but not necessarily part of it; or an Annotation with another `motivation` meaning that it is somehow about the Canvas.

A scholar who notes when a performer draws a breath or emphasizes a phrase is not really transcribing the poem--especially if the remaining text of the poem is not also represented in the associated annotations--so we're not required to use `supplementing` here.  On the other hand, I don't really understand the difference between the annotation being "from the canvas" versus "about the canvas".  

The [spec also](https://iiif.io/api/presentation/3.0/#35-values) gives some hints when it suggests behavior for presenting annotations with a `supplementing` motivation:

> Resources associated with a Canvas by an Annotation that has the `motivation` value `supplementing`  _MAY_ be presented to the user as part of the representation of the Canvas, or _MAY_ be presented in a different part of the user interface. The content can be thought of as being _from_ the Canvas. [...] For example, an Annotation that has the `motivation` value `supplementing`, a body of an Image and the target of part of the Canvas is an instruction to present that Image to the user either in the Canvas's rendering area or somewhere associated with it, and could be used to present an easier to read representation of a diagram. Similarly, a textual body is to be presented either in the targeted region of the Canvas or otherwise associated with it, and might be a transcription or translation of handwritten text or captions for what is being said in a Canvas with audio content.

When presenting these annotations, we hope to see them alongside the media.  However, they should not replace the media or be presented over the media: if, for example, these annotations were applied to a video recording of the performance instead of a pure audio recording, we would not want these annotations presented as captions.  They function as footnotes to the performance--intended to be linked to particular sections of the media--but don't replace it.  Clearly we should not use `painting` as a motivation, but does `supplementing` apply?

azaroth42:
As they're not intended to be transcription, then they're not supplementing. That said, I think the example is a poor one for a recipe due to this confusion. If the authors are confused, then people reading the recipe will be even more confused.

benwbrum:
We'll remove the paragraph about `supplementing` from the markdown, and include a sentence clarifying that non-transcription/translation content such as scholarly commentary should use `["commenting"]` motivations _without_ `supplementing`.  That might be useful for people others who--like me--found the spec's distinction between "from the canvas" and "about the canvas" unclear when implementing.

glenrobson:
Note the build is breaking as it expects the motivation to be a string... so I will need to fix that.

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: Sorry I'm wrong here it should be a string...

recipe/0103-poetry-reading-annotations/annotations.json
glenrobson: Sorry I'm wrong here it should be a string not an array...

recipe/0103-poetry-reading-annotations/manifest1.json
glenrobson: Sorry I'm wrong here it should be a string...

saracarl:
Changes from the TRC:
1)  remove the array brackets from the annotation motivation.
2)  remove the array/string discussion in the implementation notes
3)  explanation should explain "commenting" as a motivation.

Should we break this into 2 recipes; this one would only demonstrate scholarly commentary on audio files; another one would show how annotations can be internal to a manifest OR linked to an external annotation file?

benwbrum:
Changes from today's cookbook meeting:

* Fix renumbered list item in comment explanation
* Remove version with embedded annotations -- a separate recipe will discuss embedding annotations vs. externalizing them.
* Remove extra slash in URI for Harvard MP3 file.

glenrobson:
Tasks:
 * Merge in main so that you get the new styles
 * Send Glen and audio fixture for fixtures.iiif.io
 * Create a couple of annotations using Fragment selector
 * Re-work text and add links to [Start for pointSelector](https://iiif.io/api/cookbook/recipe/0015-start/) and for [remote annotations](https://iiif.io/api/cookbook/recipe/0269-embedded-or-referenced-annotations/).  
 * Only 1 manifest per recipe
 * Look at macros in: https://raw.githubusercontent.com/IIIF/cookbook-recipes/master/recipe/0001-mvm-image/index.md

glenrobson:
Related issue #103

saracarl:
- [x] Merge in main so that you get the new styles
- [x] Send Glen and audio fixture for fixtures.iiif.io -- please pull from link in manifest.   (This one we have permissions for)
- [x] Create a couple of annotations using Fragment selector -- where "a couple" is "one"
- [x] Re-work text and add links to [Start for pointSelector](https://iiif.io/api/cookbook/recipe/0015-start/) and for [remote annotations](https://iiif.io/api/cookbook/recipe/0269-embedded-or-referenced-annotations/).
- [x] Only 1 manifest per recipe
- [x] Look at macros in: https://raw.githubusercontent.com/IIIF/cookbook-recipes/master/recipe/0001-mvm-image/index.md  -- the includes at the top?  Added those.  Or the manifest/canvas references?  Did that too.

glenrobson:
List topic of options: https://github.com/IIIF/cookbook-recipes/blob/master/_data/topics.yml

benwbrum:
Changes from today's cookbook meeting:

- [ ] Only use URL fragment annotations
- [ ] Only use ranges -- leave a note in the annotation pointing to the PointSelector recipe
- [ ] Remove documentation on the verbose FragmentSelector
- [ ] Change durations to floats
- [ ] Remove all `@context` elements except for the top one, since the IIIF presentation spec inherits the web annotation spec

glenrobson:
- [ ] Remove numbered lists and turn into paragraphs 
- [ ] Remove indent for example of a fragment selector 
- [ ] "See related recipes. See [Begin playback at a specific point - Time-based media](https://preview.iiif.io/cookbook/0103-poetry/recipe/0015-start/)." Re word to say see how the point selector is used in this recpe. 
- [ ] either remove annotationpage label or make it more relevant
- [ ] duration should be a float: `"duration": 707`
- [ ] "type":"FragmentSelector",
- [ ] Add the following above the example JSON:

```
{% include manifest_links.html manifest="manifest.json" %}
```
- [ ] replace:
```
{: .line-numbers data-src="manifest2.json" }
```json
```

with:
```
{% include jsonviewer.html src="manifest.json" %}
```

glenrobson:
Glen to move sound file to fixtures.

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,86 @@
...
+Because the annotations are pointing out features of the audio, rather than transcriptions, the "motivation" for each annotation is "commenting" not "supplementing".  (If the annotations were transcriptions, their motivation would be "supplementing".)
+
+
+## Example
+
+A manifest for a poetry reading by Canadian poet Daphne Marlatt in 2018.  The recording is 707 seconds long.
+{% include manifest_links.html manifest="manifest.json" %}
+
+{% include jsonviewer.html src="manifest.json" %}
+
```
glenrobson, line 75: These link to manifest.json but your manifest is called manifest2.json. Either change these links or it maybe better just to rename your manifest to manifest.json.

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,86 @@
...
+Media Fragment using "#t=702.0,705.0" appended to the canvas URL. Instead of "source" you'd condense it into the "target" field:
+
+```
+"target":"{{ id.path }}/canvas/1#t=702.0,705.0"
+```
+
+If you are targeting a single point, you should use a point selector.  The [Begin playback at a specific point - Time-based media][0015] recipe demonstrates that approach.
+
+Because the annotations are pointing out features of the audio, rather than transcriptions, the "motivation" for each annotation is "commenting" not "supplementing".  (If the annotations were transcriptions, their motivation would be "supplementing".)
+
```
glenrobson, line 67: According to the style guide commenting and supplementing should be back ticked i.e. `commenting` not `supplementing`. also `motivations`

recipe/0103-poetry-reading-annotations/index.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Scholarly Annotation of a Poetry Reading
+id: 103
+layout: recipe
+tags: [audio, presentation, annotation]
+summary: "Use annotations to indicate aspects of the performance of a particular poem."
+viewers:
+  - Mirador
+  - Aviary Viewer
```
glenrobson, line 9: This should be Aviary not Aviary Viewer

saracarl:
Thanks for making these last update, @glenrobson

glenrobson:

## Use case
* Could you shortern and take out the discussion on tracks to match other recipes like [resrouce on a timeline](https://preview.iiif.io/cookbook/0103-poetry/recipe/0560-resources-on-a-timeline/)

## Implementation notes
* Remove example:
```
"target":"/canvas/1#t=702.0,705.0"
```
 as its not a https url. You could re-word the sentance before to “For example the Media Fragment for a duration between second 702, 705 it would look like #t=702.0,705.0“. There is the example in:https://iiif.io/api/cookbook/recipe/0485-contentstate-canvas-region/ where it does a similar description of how to use a media fragment in the section “The id of the target will contain a fragment selector “

## Example:
 * Remove Code samples:[ Python: iiif-prezi3](https://iiif-prezi.github.io/iiif-prezi3/recipes/0103-poetry-reading-annotations/)  as iiif-prezi3 doesn't support this recipe yet.

 * On viewer support the editors were stricter when defining viewer support and suggested removing the ones that don’t jump to the correct position in the audio. So remove Clover, Mirador and Theseus. 
  
With the changes above the cookbook editors are happy for it to go to the TRC.

stephenwf:
image

Could Theseus be added now? I've fixed the navigation and added timestamps under. There's room for a better user experience for sure, but it does support it now

https://theseusviewer.org/?iiif-content=https%3A%2F%2Fpreview.iiif.io%2Fcookbook%2F0103-poetry%2Frecipe%2F0103-poetry-reading-annotations%2Fmanifest.json

PR #110, opened 2019-11-04
Opera recipes

cjcolvar:
Attempt at implementing recipes #26, #64, and #65.  Use case and other notes could probably be fleshed out more but I was at a loss for words.

~Video files have been passed to Jon Dunn who will put them in the S3 bucket.  When that has happened the manifests will need to be updated and tested in a client like UV.~  

As they stand now, the manifests validate with the 3.0 validator.  I tested in the UV examples page and the three manifests playback but 64 and 65 are missing their homepage link and thumbnail images.  Additionally structural navigation isn't working for the multi canvas example.

Opera TOC: http://universalviewer.io/examples/#?c=&m=&s=&cv=&manifest=https%3A%2F%2Fraw.githubusercontent.com%2FIIIF%2Fcookbook-recipes%2F816cf76a414a2eda9f8468d83b22c6d058e55e38%2Frecipe%2F0026-toc-opera%2Fmanifest.json

Single canvas opera: http://universalviewer.io/examples/#?c=&m=&s=&cv=&manifest=https%3A%2F%2Fraw.githubusercontent.com%2FIIIF%2Fcookbook-recipes%2F816cf76a414a2eda9f8468d83b22c6d058e55e38%2Frecipe%2F0064-opera-one-canvas%2Fmanifest.json

Multi-canvas opera: http://universalviewer.io/examples/#?c=&m=&s=&cv=&rid=https%3A%2F%2Fexample.org%2Fiiif%2Fcookbook-recipes%2F0065-opera-multiple-canvases%2Fmanifest%2Frange%2Fr2f07821a-33c0-49c3-8e28-2fa3bcdf1349&manifest=https%3A%2F%2Fraw.githubusercontent.com%2FIIIF%2Fcookbook-recipes%2F816cf76a414a2eda9f8468d83b22c6d058e55e38%2Frecipe%2F0065-opera-multiple-canvases%2Fmanifest.json

Issue #65, opened 2018-12-28
An opera on multiple Canvases


tomcrane:
## An opera on multiple Canvases

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
Same opera as previous issue, now modelled as multiple canvases

kirschbombe:
Isn't this one done? Or is this a duplicate ticket?

glenrobson:
Table of Contents for Multiple A/V files on Multiple Canvases

https://iiif.io/api/cookbook/recipe/0065-opera-multiple-canvases/

Issue #64, opened 2018-12-28
An opera on one Canvas


tomcrane:
## An opera on one Canvas

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
Opera modelled as single extent of time

kirschbombe:
Done or duplicate?

glenrobson:
Table of Contents for Multiple A/V files on a Single Canvas 

https://iiif.io/api/cookbook/recipe/0064-opera-one-canvas/

Issue #26, opened 2018-12-27
Table of contents (ranges) - acts of an opera


tomcrane:
## Table of contents (ranges) - acts of an opera
(this will usually be the same as the issue name)

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
The logical structure of an audio recording - an opera, with a hierarchy of acts, arias, etc.

glenrobson:
Has this been done now?

glenrobson:
Group decided on 28th Jan to close due to it being related to #64 and #65 which have been written.

PR #136, opened 2020-01-28
Opera Recipes

cjcolvar:
Creates recipes #26, #64, and #65.  This PR is a continuation of #110.


glenrobson:
(Automated comment): Branch aviliable for testing at https://preview.iiif.io/cookbook/0026_0064_0065-opera-recipes

cjcolvar:
I think this is ready for review again.  

@glenrobson One thing I'm not sure about is if the `{{ id.url }}` and `{{ id.path }}` substitutions are actually happening in the manifests.  They don't seem to be getting replaced in the branch test site.

cjcolvar:
It appears that the branch test site doesn't have CORS enabled so I can't test out the manifests in UV.

glenrobson:
Thanks I'm looking into the CORS issue..

glenrobson:
Could you try it now? I think I've fixed it but may need to do some more CloudFront fiddling...

cjcolvar:
:tada: It's fixed now!  Thanks @glenrobson!

cjcolvar:
Testing in UV appears good for 0026 and 0064, but 0065 is having some issues.  Range highlights aren't working and playback frequently gets frozen when skipping around or switching between the canvases.  I'm not sure if those are something wrong with the manifest or UV.

Also the thumbnail images for 0064 and 0065 should get uploaded to fixtures.iiif.io, right?

recipe/0064-opera-one-canvas/manifest.json
```diff
@@ -0,0 +1,122 @@
...
+                "width": 1920,
+                "duration": 3307.22
+              }
+            }
+          ]
+        }
+      ],
+      "thumbnail": [
+        {
+          "id": "https://media.dlib.indiana.edu/master_files/vt150p78b/thumbnail",
```
cjcolvar: This thumbnail image (and the ones in 0065) should get uploaded and served from fixtures.iiif.io, right?

glenrobson: Good point. It would make sense to and include it in the same directory as the video but it will complicate my redirecting but I think its the right thing to do. 

Do you have access to the s3 bucket or would you like me to upload it?

glenrobson:
I've changed the https://fixtures.iiif.io so that it redirects videos (*.mp4 files). Does this help at all? It should mean byte requests are supported. Example redirect: 

https://fixtures.iiif.io/video/indiana/donizetti-elixir/vae0637_accessH264_low.mp4

cjcolvar:
Thanks @glenrobson.  As far as I can tell it looks like the redirect is working and byte range requests are going directly to s3 correctly.  @jwd do you want to check that it looks good to you?

glenrobson:
(Automated comment): Branch available for testing at https://preview.iiif.io/cookbook/0026_0064_0065-opera-recipes

cjcolvar:
I rebased on top of master and used the new includes to render the json viewer and manifest links.

cjcolvar:
It looks like the UV 4.0.0 prerelease on netlify isn't rendering the navigational structure for me anymore.  Maybe the UV links should be removed from recipe 26 and 64?

glenrobson:
Hi Chris, 

The cookbook editors have reviewed your recipe and have asked you to make the following changes. I've grouped them by individual recipe. 

[Preview #26](https://preview.iiif.io/cookbook/0026_0064_0065-opera-recipes/recipe/0026-toc-opera/) - Table of contents (ranges) - acts of an opera
 * Could we rename the recipe: `Table of contents for A/V content` to make it more generic? 
 * Can we have more context in the use case? The other recipes had more detail in the Use case section and we were left wondering is this a book of an opera a video or recording etc. It also assumes you know what an opera is and how it would be broken up.  Maybe a little note about what opera, act and aria are...
 * Can you add the `Multiple languages in metadata` recipe as a related recipe?
 * Could you expand on the Overlapping ranges note in the restrictions? Maybe with an example?

[Preview #64](https://preview.iiif.io/cookbook/0026_0064_0065-opera-recipes/recipe/0064-opera-one-canvas/) - An opera on one Canvas
 * Can you add the thumbnail and multi-langauge metadata recipe to Related recipes?
 * We think the annotation of the videos on the canvas is wrong, and they should be targeting #t=0 for the first one, and #t=3971.24 for the second
 * Once corrected could you highlight this detail in the implementation note?
 * Could you go into more detail on when would you want to use more than one canvas? Is this to achieve the "continuous playback"?
 * Table of contents not showing in UV, is this a UV issue?
 * Can you think about the title of the recipe maybe something on the lines of“Table of contents on a single canvas for A/V content” to make it more generic as the previous one. 

[Preview #65](https://preview.iiif.io/cookbook/0026_0064_0065-opera-recipes/recipe/0065-opera-multiple-canvases/) - An opera on multiple Canvases
 * Similar to #64 title: “Table of contents on multiple canvases for A/V content”?
 * Does this reciepe work with the UV? We didn't see a UV link.
 * Can you add metadata and thumbnail and basic video to the related recipes?
 * Do we want to call out segmented playback functionality? Is this a restriction?

recipe/0064-opera-one-canvas/manifest.json
```diff
@@ -0,0 +1,122 @@
...
+        },
+        {
+          "id": "{{ id.path }}/canvas/1/annotation_page/1",
+          "type": "AnnotationPage",
+          "items": [
+            {
+              "id": "{{ id.path }}/canvas/1/annotation_page/1/annotation/2",
+              "type": "Annotation",
+              "motivation": "painting",
+              "target": "{{ id.path }}/canvas/1",
```
glenrobson, line 52: I think this should have `#t=3971.24` on the end to make sure the audio is target to the correct place on the canvas.

recipe/0064-opera-one-canvas/manifest.json
```diff
@@ -0,0 +1,122 @@
...
+      "items": [
+        {
+          "id": "{{ id.path }}/canvas/1/annotation_page/1",
+          "type": "AnnotationPage",
+          "items": [
+            {
+              "id": "{{ id.path }}/canvas/1/annotation_page/1/annotation/1",
+              "type": "Annotation",
+              "motivation": "painting",
+              "target": "{{ id.path }}/canvas/1",
```
glenrobson, line 32: This may not be strictly necessary but as your targeting a specific time in the second canvas I think it would be more explicit if you add `#t=0` on the end of this target.

glenrobson:
Changes requested by TRC have been done and cookbook group reviewed them.

PR #141, opened 2020-03-09
Update index.md

cubap:
I tried to just simplify in most cases. I added some suggestions for `<code>` wrapping property names to call them out, which I think is conventional here.
I changed the example Point coordinates to make it more obvious that it is a latlng number and not something like a Canvas x,y.
Take or leave whatever.


thehabes:
This page has been redone and this will cause way too many conflicts.  If you want to implement these changes, do them in the .json files.

cubap:
`48.86, 2.34` is the coordinates that I was using in the Point and `service` block URIs.

Issue #13, opened 2018-12-27
Poster Image Displayed Before Interaction


tomcrane:
## Poster image displayed before interaction

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I wish to provide the user with something to look at before they choose to start interacting with the object, and/or while they wait for it to load/buffer.

tomcrane:
Discussed in WG meeting - this is the same modelling as for journals or newspapers, but it's worth producing a separate recipe for a journal splash / poster image. It will make it easier for people to find, and copy.

triplingual:
Note to self: working example (more or less) of this idea in v2 at http://universalviewer.io/examples/?manifest=http://wellcomelibrary.org/iiif/b17307922/manifest#?c=&m=&s=&cv=&manifest=http%3A%2F%2Fwellcomelibrary.org%2Fiiif%2Fb17307922%2Fmanifest and https://uv.netlify.app/#?c=&m=&s=&cv=&manifest=http%3A%2F%2Fwellcomelibrary.org%2Fiiif%2Fb17307922%2Fmanifest (note the difference in placeholder image sizes)

PR #151, opened 2020-04-13
Recipe #0013 — Poster Image Displayed Before Interaction

triplingual:
This PR addresses issue #13, showing something (a `placeholderCanvas`) to the user to view before they choose to start interacting with the object and/or while they wait for it to load/buffer.


glenrobson:
(Automated comment): Branch available for testing at https://preview.iiif.io/cookbook/0013-placeholderCanvas

glenrobson:
I asked on the UV slack and it looks like it implements `posterCanvas` [here](https://github.com/UniversalViewer/universalviewer/blob/2a08b8aa87dce7dd5d4d6c65c63df495bebe9775/src/lib/manifold.js#L764) which was the original name for the field before it was renamed `placeholderCanvas` [here](https://github.com/IIIF/api/issues/1605#issuecomment-417031935).

glenrobson:
Comments from Cookbook editors:

They are happy to send it to the TRC with the following changes:

 * Use case: can you make the use case more specific. Talk about using a large video and and showing a image while it loads.
 * Could you change "Note: The placeholderCanvas property for time-based objects is not yet supported in viewers." to "Note: The placeholderCanvas property is not yet supported in viewers." - so as not to give the impression that it works for Image content. 
* We found one contraction: "Don’t use this property for content"

PR #150, opened 2020-04-09
Recipe #0009 — Simplest Manifest - Multiple Related Images (book, etc.)

regisrob:
This PR is a first pass at addressing issue #9, creating recipe 9 : simplest manifest with multiple related Images (book, etc.)

TODO's:

- [x] add a sentence about the recommended use of `thumbnail` property into the implementations notes
- [x]  find a few sample images in Gallica or elsewhere (pages of a book...)
- [x] update the manifest: fixtures image urls, image and canvas sizes, canvas labels
- [x] add links to viewers
- [ ] check the viewers once the manifest is accessible via http

PR #153, opened 2020-04-20
Recipe #0009 — Simplest Manifest - Multiple Related Images (book, etc.)

glenrobson:
Importing @regisrob recipe from https://github.com/IIIF/cookbook-recipes/pull/150


glenrobson:
Recipe available at: https://preview.iiif.io/cookbook/0009-book-1/recipe/0009-book-1/

glenrobson:
1 small change. Would it be possible to add the width and the height for the canvas thumbnails? In the spec in the [thumbnail section](https://iiif.io/api/presentation/3.0/#thumbnail) it says:

"Images and videos should have the width and height properties, and time-based media should have the duration property" 

and I think for clients to make use of the thumbnail they would probably need the dimensions.

regisrob:
@glenrobson Yes, I will add them ;)

glenrobson:
(Automated comment): Branch available for testing at https://preview.iiif.io/cookbook/0009-book-1

glenrobson:
Hi Régis,

The cookbook editors reviewed your recipe and wondered if you could make the following changes:

 * Move 2nd paragraph from Use case to Implementation note (the one starting with "The sample Manifest below represents")
 * Can you use a IIIF image service in canvses rather than flat images?
 * Can you add the following related recipes to thumbnail recipe (thumbnail algorithm / discussion) and to Simple Image Service (Image Service for single image (24,25))
 * Add in the viewer include into the recipe and we note the UV is currently not working but by moving to an image service that might avoid the current known issue with flat images on fixtures not setting the correct CORS information. 
 * At the end of the recipe there are three ``` symbols 

Thanks

regisrob:
Hi Glen, thanks for the feedback, I will make the proposed changes. I was already thinking of adding the image `service`s (partly because M3 seems to require it on the `thumbnail`). I may have fixed the ``` symbols already but I will double check.

regisrob:
Hi @glenrobson, this point is unclear to me: "Can you add the following related recipes to thumbnail recipe (thumbnail algorithm / discussion) and to Simple Image Service (Image Service for single image (24,25))".
Should I just point to these two recipes from the “Related recipes” section of 0009-book-1?
Thanks

glenrobson:
Yes, that was the intention. So just add the thumbnail recipe (thumbnail algorithm / discussion) and to Simple Image Service (Image Service for single image (24,25)) below the `Related recipes` heading.

regisrob:
Done via #161.
Anticipating a little bit on potential feedback in the next phase: I realized that because the first page in this sample book is not supposed to be presented as a single page (it is a frontispiece that should face the title page) the "two page view" (UV) or "book view" (Mirador) are currently "broken". I guess adding at least one canvas (or 3) at the beginning of the sequence should be enough to solve the problem...

glenrobson:
Thanks Régis, we discussed this in the cookbook meeting and thought adding another image to make it even would clear up any confusion with the UV viewing so if you do have another image send it to me and I can upload it.

glenrobson:
Looks good to me.

regisrob:
> Looks good to me.

Thanks Glen. Let me know if there are still improvements to make following yesterday's call.

regisrob:
[Last commit](https://github.com/IIIF/cookbook-recipes/pull/153/commits/733b17af34b819b39349516b800f08aa383231ff) addresses the remarks from the TRC call, so ready for new review.

_includes/links.md
```diff
@@ -4,8 +4,11 @@
 [0002]: {{site.baseurl}}/recipe/0002-mvm-audio/ "Simplest Manifest - Audio"
 [0003]: {{site.baseurl}}/recipe/0003-mvm-video/ "Simplest Manifest - Video"
 [0007]: {{site.baseurl}}/recipe/0007-string-formats/ "Embedding HTML"
+[0009]: {{site.baseurl}}/recipe/0009-book-1/ "Multiple Related Images (Book, etc.)"
```
glenrobson, line 7: Could you update the title to "Simple Manifest - Book"

index.md
```diff
@@ -33,7 +33,7 @@ _The corresponding 2.1 test fixture(s) is given like this, where appropriate: ..
 * [Embedding HTML in descriptive properties][0007] (64)
 * Rights statement(s) (7)
 * thumbnail algorithm / discussion
-* Book (simplest, > 1 canvas) (19)
+* [Multiple Related Images (Book, etc.)][0009] (19)
```
glenrobson, line 36: Could you update the title to "Simple Manifest - Book"

_includes/links.md
regisrob: Done in #209

index.md
regisrob: Done in #209

glenrobson:
Changes requested by TRC made and approved by cookbook group.

PR #168, opened 2020-05-20
Deploying master to iiif.io

glenrobson:


glenrobson:
Moving to master and merging.

Issue #14, opened 2018-12-27
Additional optional content accompanying the thing being modeled


tomcrane:
## Additional optional content accompanying the thing being modeled

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I wish to provide the user with something to look at while they listen to the work represented by the IIIF resource, or I wish to provide something for the user to listen to while they interact with the visual work represented by the IIIF resource.

Important that the additional content can be ignored without loss of fidelity of the modeled thing.

Example cases to make recipes:
  * Manuscript with an (unrelated) time-appropriate music to play in the background
  * Ads to show while a podcast playing

jazairi:
Discussed at working meeting. Examples added for clarification.

triplingual:
A real-world example of the application of this issue came to my notice this morning: [Playing Images: “The Silver Moon”](https://artgallery.yale.edu/article/playing-images-silver-moon) It's the COVID-19 forced-online version of [art+music talks that are given in-person at the YUAG regularly](https://artgallery.yale.edu/overall-search/%22playing%20images%22).

The Yale University Art Gallery doesn't use IIIF for image display, but they could! 😀 And it will be good to have this recipe available for them.

PR #172, opened 2020-05-29
Recipe #0014 — accompanyingCanvas

triplingual:
This PR addresses issue #14, additional optional content accompanying the thing being modeled. Note that the additional content must be ignorable without loss of fidelity of the modeled thing. Examples include a text document as the thing with music to play in the background.


glenrobson:
The cookbook editors approved this recipe ready for TRC with one minor change. Would it be possible to remove the backticks from Canvas from the following:

“main `Canvas` contains audio”

Under the Example heading.

triplingual:
Oh -- I had a note to self in my local copy to make the intro use case a little more specific just as #13 needed more specificity there. I'll let the TRC weigh in on that, then?

glenrobson:
Approved by TRC so merging.

Issue #139, opened 2020-03-05
Place Canvas Fragment to an Area on a Web Map


thehabes:
## Geolocating A Canvas Fragment

- [index.md File](https://github.com/IIIF/cookbook-recipes/blob/0139-geo-annotation/recipe/0139-geolocate-canvas-fragment/index.md)
- [Preview](https://preview.iiif.io/cookbook/0139-geo-annotation/recipe/0139-geolocate-canvas-fragment/)

## Use case  
SLU is using geographic Annotations to describe JSON-LD resources.  We use mapping tools like Leaflet to render these Annotations into meaningful interfaces.  The general use case is I know the JSON-LD URI of some data resource.  I may own the database in which it presides, I may own the resource, or it may be entirely third party.  The resource might not be IIIF, but for our purposes here say that it is.  I have geographic information, which may include some metadata beyond geographic coordinates, that applies to this resource.  I would like to make a IIIF compliant assertion that places the resource to known Earth coordinates.   

## Further Information
This is originating from the IIIF + Maps conference, the original issue can be found at https://github.com/IIIF/iiif-stories/issues/118.

The recipe is a stepping stone to recipes for the following stories
- https://github.com/IIIF/iiif-stories/issues/116
- https://github.com/IIIF/iiif-stories/issues/119
- https://github.com/IIIF/iiif-stories/issues/125
- https://github.com/IIIF/iiif-stories/issues/135

Recipes for those stories should note this recipe and, where possible, will be collapsed into this recipe.

thehabes:
There are multiple ways to do this depending on your situation.  If you own the resource, you can either add a `service` block containing the geocoding information to the resource directly, or you can choose to annotate the resource.

If you choose to annotate, here is what that looks like in Presentation API 3
```JSON
{
   "id":"https://example.com/annotation/12345",
   "type":"Annotation",
   "@context":"http://iiif.io/api/presentation/3/context.json",
   "motivation":"geocode",
   "body":{
      "id":"https://example.org/geojson/id/123",
      "@context":"http://geojson.org/geojson-ld/geojson-context.jsonld",
      "type":"Feature",
      "properties":{
         "label":"String this geometry node should use as a label.",
         "description":"Some description to go along with node."
      },
      "geometry":{
         "type":"Point",
         "coordinates":[
            7,
            7
         ]
      }
   },
   "target":"https://iiif.example.org/canvas/123"
}
```

[...]

PR #183, opened 2020-06-26
Represent Canvas Fragment as a Geographic Area on a Web Map

thehabes:
Files are at https://github.com/IIIF/cookbook-recipes/tree/0139-geo-annotation/recipe/0139-geolocate-canvas-fragment

Preview : https://preview.iiif.io/cookbook/0139-geo-annotation/recipe/0139-geolocate-canvas-fragment/

Issue that started this recipe is at https://github.com/IIIF/cookbook-recipes/issues/139

I am willing to bet the recipe ID (index and the /recipe folder name) are off, please let me know what number you actually want me to make it.


thehabes:
I am willing to bet the recipe ID and stuff are off, please let me know what number you actually want me to make it.

recipe/0139-geolocate-canvas-fragment/index.md
```diff
@@ -0,0 +1,32 @@
...
+
+[`geometry` can be more than just a `Point`.](https://tools.ietf.org/html/rfc7946#section-3.1)
+
+### Restrictions
+Nested GeoJSON coordinate arrays are incompatible with the processing model of JSON-LD 1.0. The JSON-LD 1.1 processing model does not have this restriction.
+
+### Example
+An example of a Manifest that has a Canvas fragment geolocated to a geographic point. The Manifest contains one Canvas with one Image, and the Canvas has the same size dimensions as the Image. The Canvas contains one Annotation Page with one Annotation targeting the region of interest where "Paris" appears using the [#xywh Fragment Selector syntax](https://www.w3.org/TR/annotation-model/#fragment-selector). Since the image used is a IIIF Fixture following [IIIF Image API 3.0](https://iiif.io/api/image/3.0/), you can see the targeted fragment by supplying [the values used in the #xywh selector to the image URL](https://iiif.io/api/image/3.0/example/reference/59d09e6773341f28ea166e9f3c1e674f-gallica_ark_12148_bpt6k1526005v_f20/1300,3370,250,100/max/0/default.jpg). 
+
```
glenrobson, line 24: Could you add a link to the manifest by including the following here:

{% include manifest_links.html viewers="" manifest="manifest.json" %}

glenrobson:
Feedback from Editors:

### In Implementation Notes
 * Could you reword the following sentence:
“If, for example, the targeted resource has a label and the properties field contains a label, the consuming interface must make a choice on which to preference for presentation purposes.”

to change "which to preference for presentation purposes" to “which to prioritize for presentation purposes.”
 * Could you reword:
“geometry can be more than just a Point.” 

to "Note that geometry can be more than just a Point."

### in Restrictions
 * Could you add something to say that most people won't need to worry about the 1.0 and 1.1 restriction. Something like for those that are interested in converting the JSON-LD to RDF should be aware that...  

### In Example
 * Missing link to manifest. See inline review above but it is the following snip it above the json viewer:

```{% include manifest_links.html viewers="UV, Mirador" manifest="manifest-individuals.json" %}```
 
### In Related recipe
 * Add a new recipe for feature collection (this would be a new recipe yet to be written that would link a canvas fragment to a feature collection rather than a feature). 
 * Add a link to the Fragment selector recipe
 * Add a link to the Tagging annotation recipe

The cookbook editors are happy for you to make the changes and then for the recipe to go to TRC.

File: _includes/links.md [hunk 2]
```diff
@@ -13,6 +13,7 @@
 [0202]: {{site.baseurl}}/recipe/0202-start-canvas/ "Load Manifest Beginning with a Specific Canvas"
 [0014]: {{site.baseurl}}/recipe/0014-accompanyingcanvas/ "Audio Presentation with Accompanying Image"
 [0015]: {{site.baseurl}}/recipe/0015-start/ "Begin playback at a specific point - Time-based media"
+[0139]: {{site.baseurl}}/recipe/0139-geo-annotation/ "Geolocate Canvas Fragment to a Point"
```
glenrobson, line 16: I think this link should be: recipe/0139-geolocate-canvas-fragment/

thehabes: OOO man I see.  I thought this was going off the branch name, not the branch/recipe name.

glenrobson:
Unfortunately there is some work to do on this recipe before it can go back for approval at the TRC. The issues raised are in this ticket:

https://github.com/IIIF/trc/issues/51

glenrobson:
Hi Bryan,

The editors reviewed your recipe and are generally happy for it to go to the TRC but have asked for a quick review once you've made the requested changes below:

* Could you change the:

“to associate this map with modern geocoordinates that clients can use for representing the targeted region in world based Web Maps, such as Leaflet or Google Maps”  

to 

"to associate this map with a geographic coordinates system. Once associated clients could use geographic coordinates to represent the targeted region in clients like Leaflet or Google Maps. "

The reason was to make this more approachable to non Geo people. They also requested to remove reference to Web Map just to avoid confusion or conflation by mentioning another Standard. 
 * Move to restrictions: “When the @context property is used as an array with multiple contexts, the IIIF Presentation API 3.0 context must be the last item in the array.”  
 * Could you replace the TBD  in the recipes under “Related Recipes” with issue numbers? If they don't have issue numbers could you remove the related recipe. 
 
They were happy with the switch of use case and thought it made it a lot clearer. 

Thanks

Glen

thehabes:
**Response to the first change suggestion**
You are not just trying to associate this map with a geographic coordinates system.  If that's what you're doing, GeoJSON may not be the best option.  You are trying to associate the map (on a fragment of this IIIF Web resource) with WGS84 compliant coordinates for use on the Web in hopes you could view those coordinates rendered in some way on an online (Web) map like Google Maps (and maintain the connection to metadata about the IIIF Web resource).  Online Web map coordinate stuff is done with GeoJSON, and any data you have that is not GeoJSON is being shimmed into GeoJSON for this purpose, which is the argument for minting those objects with GeoJSON as opposed to anything else.  

As such, I can suggest something where we meet half way

- to associate this map with WGS84 compliant coordinates for use in online mapping systems (a.k.a Web Maps).  GeoJSON is the primary data format processed by online mapping systems (Web Maps), and so the coordinates for the targeted region can be processed by clients (Web Maps) such as Leaflet or Google Maps

**Response to second change suggestion**
- I think Web Map is the most clear pronoun.  `Online Mapping System`, `Coordinate Reference System`, `MapUI`, `Map-Based Web Application` all seem much more vague to me.  Web Map has become accepted vocabulary.  I am reluctant to change it without very good reasoning to the contrary.  

**Response to third change**
The JSON-LD 1.1 spec notes `Multiple contexts may be combined using an array, which is processed in order.` and places no restriction on order, since it is a `@set`.  The IIIF Presentation API 3 restricts the freedom of where you can place the IIIF Presentation API 3 context because of its implementation demands, but otherwise places no restrictions here.  These feels more like a (IIIF Presentation API 3) implementation note than a restriction, and I am reluctant to move it without more reasoning.  

Sorry to be so confrontational!  I am really a reasonable person and am happy to talk further about any of this.

eliotjordan:
I would suggest:

"to associate this map with WGS84 compliant coordinates for use in web map mapping clients like Leaflet and OpenLayers."

thehabes:
**Post IIIF Maps TSG Call Feelings**
- See Elliot's new wording suggestion.  Important to let people know this is for connecting with familiar web mapping clients like Leaflet, Google Maps, OpenLayers, OpenStreetMaps (pick just a couple).  Not just for assigning coordinates w/o a presentation purpose.  
- Capitalized `Web Maps` is a main component of causing these feelings.  At least lowercase it.  No WGS84 that's too technical.  General agreement on this change.
- Leave it as an implementation note.

Thanks to everyone on the call today.  It was extremely helpful for thinking through this.  I will get these changes in.

glenrobson:
Thanks Bryan its looking good!

PR #189, opened 2020-07-10
Represent Manifest as a Geographic Point on a Web Map

thehabes:
Preview : https://preview.iiif.io/cookbook/0182-geolocated-simple-manifest/recipe/0182-geolocated-simple-manifest/


thehabes:
Closing this.  We are now accomplishing this through different recipes using the `navPlace` API extension.

PR #190, opened 2020-07-13
Americanized spellings 🇺🇸

triplingual:
Both the spec and the emerging style guide for the cookbook are in American English, so I changed places where words used British English spellings.


glenrobson (approved):
Looks good to me.

PR #165, opened 2020-05-15
5. Introduce IIIF Image service

glenrobson:
Starting this off to see how much work is required.

PR #197, opened 2020-08-05
Recipe #0005 — Image Service for single image

triplingual:
This PR addresses issue #5, with reference to [the PR for the old name of the branch](https://github.com/IIIF/cookbook-recipes/pull/165). This is a minimum viable manifest that uses an Image service. All subsequent recipes that use an Image service should be able to refer to this one as a starting point.


triplingual:
Preview main page: https://preview.iiif.io/cookbook/0005-image-svc-single-image/
Direct link to recipe: https://preview.iiif.io/cookbook/0005-image-svc-single-image/recipe/0005-image-service/

glenrobson:
The cookbook editors have the following requests for changes but are happy for it to go to TRC when incorporated:

In implementation notes could you highlight the other features of the image api e.g. rotation, color choice? and being able to cut images for annotation. Potentially with an example URL to show what is possible? This adds to the sales pitch in the use case.  

Related recipes:
Regarding the related recipes it was though it was better to limit the number of linked recipes. As this is one of the foundation use cases if you link to all relevant recipes the list could get quite long. So it was suggested to just leave the following:
 * Presentation version 3 manifest containing images on a version 2 service
 * Services
 * Simple Manifest
 * Thumbnail 

In the last paragraph of the implementation notes, we wondered if the following sentance could be reworded from:

“Depending on your IIIF Image Service’s stabilty,” to “As an optimisation you may want to include all of the info.json content (excluding the @context) as the content of the service sections of your manifest.” 

Could you link the first mention of the IIIF Image API to the Specification website?

triplingual:
(Munging into a to-do list)

- [ ] Highlight other features of the Image API, e.g. rotation, color choice, and being able to cut images for annotation
- [ ] (if possible) Add example URL to show capabilities of image in Image API
- [ ] Limit linked recipes to: Presentation version 3 manifest containing images on a version 2 service, Services, Simple Manifest, Thumbnail
- [ ] In "Implementation Notes" section, reword “Depending on your IIIF Image Service’s stability,” to “As an optimisation you may want to include all of the info.json content (excluding the @context) as the content of the service sections of your manifest.”
- [ ] Link first mention of the IIIF Image API to its website

triplingual:
For adding an example URL that shows the capabilities of an image with the Image API, this is a bit of choice paralysis. On the one hand, there are millions of images out there using V2, but even if we want this to be an image using a V3 service, it's a little difficult to pick from among all the choices.

Would the goal be to find something that has all the affordances listed in the use case? ("proper deep zooming of large high-resolution images, client generation of derivatives, annotation of and external reference to image fragments, image rotation, choice of colors, creation of image fragments for annotation")

glenrobson:
Thanks Trip changes look great.

glenrobson:
Approved by TRC with no changes so merging.

Issue #11, opened 2018-12-27
Book (paging variations)


tomcrane:
## Book (paging variations)

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I have objects that have the concept of recto and verso; are paged. I have objects that have no such concept. I have objects that are photographed in different ways (one image per page, oneimage per pair of open pages, etc).

tomcrane:
Equivalent to 2.1 fixtures

https://iiif.io/api/presentation/2.1/example/fixtures/15/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/16/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/17/manifest.json

PR #198, opened 2020-08-06
0011 book 3 behavior

kirschbombe:


triplingual:
Establishing link with issue #11

glenrobson:
Cookbook editor comments:

The cookbook editors note this recipes contains two use cases in one and this may come up in the TRC although the editors are happy to approve it as all use cases match the feature described. 

### In use case
 * Can you mention `unordered` in the use case and the situations where this might be used. The recipe mentions ` individuals, continuous and paged` but not `unordered`. A example where it might be used is a collection of unordered photographs. Its fine to add it as a use case with context but not include an example manifest. 

### In Restrictions
 * Could you remove the back ticks around Collection, Manifest, Range and Canvas in the Restrictions section. 

### in Related recipes:
 * Can you remove Book (paging variations) - as this seems to be a link back to itself. 
 * Remove [Thumbnails][0012]
 * Remove [Table of contents (ranges) - book chapters][0024]

The cookbook editors are happy to approve the recipe with the changes above.

kirschbombe:
Thank you for the feedback. I'm happy to make the changes above, but have a question about the use case. In discussing the use cases for this recipe, we decided to only include continuous and individuals because the recipe was for behavior variations for book objects. Unordered doesn't fit this use case, so we left it out. I think we'd need to change the focus of the recipe overall to include unordered. I happy to do so, but want to make sure that is the intention. I could also create a new recipe that focuses on behavior for something like an archival folder.

glenrobson:
Thats a good point. The thing that came up in the meeting was that `unordeded` is mentioned in the restrictions part but no where else. Is there a way to include a reference to it in the use case even if only briefly and maybe point to an unordered / archival recipe?

glenrobson:
Issue passed in TRC with no changes requested so I am merging.

Issue #6, opened 2018-12-27
Multiple languages / Internationalization 


tomcrane:
## Multiple values and languages

https://github.com/IIIF/cookbook-recipes/blob/master/cookbook/0006-text-language/index.md

## Use case  
If you have more than one language for descriptive properties like metadata, label, description etc. This will show single language and multiple language.

tomcrane:
Equivalent to 2.1 fixtures

https://iiif.io/api/presentation/2.1/example/fixtures/3/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/4/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/6/manifest.json

tomcrane:
This issue could also be used to make it clear that the way languages work in IIIF and the way they work in the anno model are different.

Or is that worth a separate issue to really highlight the point?

The spec currently says (in 4.4, which introduces language maps) 

> Note that this does not apply to embedded textual bodies in Annotations, which use the Web Annotation pattern of value and language as separate properties.

mattmcgrattan:
This doesn't just apply to labels, it can also apply to other textual fields in the manifest.

mattmcgrattan:
No reference to common search terms like Internationalisation, and doesn't make clear that there are two issues involved:

* multiple values (e.g. multiple English labels for a resource)
* multiple languages (e.g. French, English, and Japanese labels for a resource)

mattmcgrattan:
Change to Multiple languages (Internationalism), and make a new issue for Multiple Values

jrgriffiniii:
This recipe has been moved to https://github.com/IIIF/cookbook-recipes/blob/master/recipe/0006-text-language/index.md

kirschbombe:
The previous recipe iteration covered both multi-value and multi-language. The cookbook team decided to break these into separate recipes: multi-value; multi-lingual; and a third for specifying region/script

PR #215, opened 2020-11-04
0006 internationalization and multi-language values

kirschbombe:
Issue: https://github.com/IIIF/cookbook-recipes/issues/6
Preview: https://preview.iiif.io/cookbook/0006-multilingual/recipe/0006-text-language/


glenrobson:

### Implementation notes
 * Could you move the last two paragraphs to the front as they seem very important messages for the reader. 
 * Could you re-write the dot notation message to:

"It should be noted that BCP 47 allows hyphen breaks for locale and scripts (e.g. en-US) and if you are processing these manifests with Javascript it will break the . notation features for navigating JSON object so you should use brackets for accessing language properties."

### Example
Could you change: “The image was sourced via Wikimedia Commons and is public domain.” to “The image in this example was sourced via Wikimedia Commons and is public domain.”

### Related recipes

Could you remove the existing ones as they don't seem awfully relevant and add a link to the rights and metadata recipes? 

Other than the above the editors were happy for this to go to TRC without another review.

kirschbombe:
Thanks, @glenrobson. I made the requested changes, but had to rework the Implementation notes section a bit since just moving the two paragraphs to the top affected the flow. Can you review the edits to make sure they look OK and achieve the desired result?

glenrobson:
It looks great Dawn. Thanks

glenrobson:
This has been approved by the TRC but they have requested the following changes:

* Remove the none case from the label in example lines 12-14 
* It would be useful to have an example of a metadata element with both labels and values in multiple languages (French and English as per the current examples)

kirschbombe:
@glenrobson - I removed the `none` case from the manifest label and I updated the `summary` property with both English and French. I also changed the field in the `metadata` property to "Subject" and included subject values in both English and French. This should take care of the fixes, but let me know if it still needs work.

glenrobson:
Sorry Dawn would it be possible to include the subject example in addition to the author one? I think its useful to show both and I believe the intention of the comment from TRC was in addition rather than replacing.

kirschbombe:
@glenrobson I can add the author back in, but I can't really make that one both French and English which is why I switched it to Subject. It will still use "none" for the value language map -- will that be OK?

glenrobson:
Thanks Dawn thats perfect. I think its important to show what I think will be a common use case that you have translated labels but not content.

PR #223, opened 2021-01-07
0219 using caption file

mcwhitaker:
First Draft.


glenrobson:
Just for completeness here are the comments from the Cookbook editors:

 * Add a label for the VTT file in manifest
 * Mirador issue is now fixed in dev version: https://deploy-preview-3379--mirador-dev.netlify.app/__tests__/integration/mirador/video.html so its OK to include Mirador in the link.
 *  Remove UV as this currently isn't working.
 * May need to remove the numbers that annotate the subtitles in the VTT file although the file does validate. 
 * Related recipes    
   * Add: seeAlso and rendering 
   * Add: Transcript - this would be an alternative rendering 
 * In implementation note can you explain that a transcription which is a text based version of the resource would be a rendering and captions are x and live in y. 
 * Potentially rename title to: “Caption and subtitles with your video content”. 
 * Restrictions: 
    * Add note the effect of: "VTT was chosen due to its wide support in browsers. Other formats are supported in IIIF but may not display in viewers. "

Overall decision was this was ready for TRC with the above comments incorporated.

glenrobson:
Hi Maria,

I've checked the changes requested and this looks good to go apart from the following change:

 * Add: seeAlso and rendering to related recipes. 

I've also merged in a recipe that was approved from the last TRC so if you could update from Master that would be great. 

Thanks 

Glen

glenrobson:
For future reference `rendering` is listed as a related recipe and Maria and I discussed seeAlso and as its not in the recipe it doesn't seem to make sense to add it as a related recipe.

glenrobson:
Recipe approved by TRC with only Simeon's comment to address:

https://github.com/IIIF/trc/issues/59#issuecomment-763839786

Note there is ongoing discussion in the IIIF #AV slack on a few extra issues but these were outside of the timeline for the TRC review. I suggest we merge this recipe and create an issue for the points in the #av discussion and then handle the changes like we have with previous recipes. If its minor we can make the change and inform TRC, if its more involved take it for another round of approval. I suspect its the second option with the discussion but we'll see how it resolves.

glenrobson:
Thanks Maria, those changes look great. I'm going to merge this now.

PR #228, opened 2021-01-16
0030-multi-volume

kirschbombe:
Preview: https://preview.iiif.io/cookbook/0030-multi-volume/recipe/0030-multi-volume/


glenrobson:
Comments from the Authors meeting on the 26th of Feb:

 * Restrictions: remove straight brackets but we like the restrictions 
 * “An alternative approach might be to combine the volumes into a single Manifest and use Ranges to create a table of contents for navigating the multiple volumes. For this approach, consult the [Book with Table of Contents][0024] recipe.” - should we only supply one option or is there a use case for doing it this way instead? Is this because its bound in the same volume? 
 * Rename collection to collection.json 
 * Add JSON-LD | View in Universal Viewer | View in Mirador above each manifest (v1 / v2)
 * Related recipes:
    * table of contents
    * Bound multi-volume work
    * Paged collections
    * Archival collection?
 * Metadata - could you investigate if we should make a recommendation on which level to put the metadata? Should it be at collection and repeated in the manifest or does it not matter for the split volume use case?

kirschbombe:
@glenrobson - I made the suggested edits, but I'm unsure about the metadata note and whether it's useful or hitting the right note. Let me know what you think, or we can just wait to see what the Editor group thinks :)

glenrobson:
Thanks Dawn, Ill make sure we discussed it in the authors meeting.

glenrobson:
Cookbook editor comments:

 * “specific use case or context; For example” - regarding the `;` in the sentence could you either replace with a full stop or use a lower case F in For example.
* In Example could you point out the manifest has a behaviour of `individuals`which isn’t necessary for this recipe but shows the two behaviours working together which we thought was useful and link to the behaviours recipe. Alternatively you could also not specify the behaviour in the manifests as `individuals` is the default. 
 * `"id"` in collection manifest still says manifest.json rather than collection.json 
 * Json-ld links point to manifest.json rather than manifest-v1.json and manifest-v2.json. 
 * When linking to the table of contents recipe in the implementation notes could you replace this link with a link to Bound Multi-volume Work that is mentioned in related recipes. 

The cookbook editors are happy for this to go ahead to TRC.

kirschbombe:
I've incorporated the changes requested by the editors. I opted to leave the behavior property in the manifests since the comments said that it was useful. If anyone wants to review that section to make sure I hit the right note, please do!

glenrobson:
Approved by TRC but they requested the following changes:

"Could we look at the title and maybe rename it something like "Multi-volume work, unbound volumes" to distinguish it from the other recipe on Multi-volume bound volumes.

Layout Question
3 JSON code blocks, but might be helpful to call out clearly what each code block is."

glenrobson:
Thanks Dawn, looks good. Merging.

Issue #230, opened 2021-02-12
Chronology with navDate


triplingual:
## Chronology with `navDate`

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
The recipe user has one or more periodical articles (journal, magazine, etc.) in IIIF form and would like to communicate their dates to a consuming client for the client to put in a chronological order.

glenrobson:
Just for completeness here is the notes from the last cookbook call:

 * Simple manifests better
 * Probably need to have multiple Canvases
 * UV will allow browsing by date, but only on a Collection of Manifests
 * Unclear which viewers support and in which way
 * Can link to newspapers recipe in Related for an example of a run of titles
 * Could use something from https://www.loc.gov/maps/?q=Chesapeake+and+Ohio+Canal&fa=contributor%3Aunited+states.+national+park+service&st=list&c=100
    * But use one and mention in Example section that there are others
    * Item fixtures link https://fixtures.iiif.io/info.html?file=/images/loc/chesapeake_map/88695674.jpg

PR #231, opened 2021-02-12
Recipe #0230—Chronology with navDate

triplingual:
This PR addresses issue #230, introducing the way to communicate date information to a consuming client. This property and its value are not intended to be displayed to end viewers.


glenrobson:
Hi Trip,

I'm afraid the cookbook editors had difficulty with this one. Nothing to do with the way the recipe was written, as all agreed that the writing looked good. The difficulty was in the example and the fact it doesn't work in viewers currently and where the navDate is currently located in the canvas. We could think of a few use cases where navDate could be useful in a canvas but so far nobody has pushed for this to be implemented in any of the viewers and so there was discussion on whether this recipe would be useful to drive implementation or if it would be a use case which is spec complaint but less likely to be one people would use. There was also discussion on whether the canvas navDate use cases would be better served by a navDate in a range.

To avoid this issue of canvas navDate would you be OK to move to a collection with 2 manifests all with navDates (at manifest level in the manifests and in the manifest object in the collection)? I can upload another LOC image to fixtures. This will hopefully work with the UV and it will be a useful recipe that focuses on just the navDate property rather than the multiple newspaper related features in the Newspaper recipe. Does this sound OK?

A couple of other notes:
 * Chronology with navDate - could the title be changed to something like "Navigating in chronological order" so as not to have the property name in the title. 
 * Could you cut down info.json of the image to just the `id`, `type` and `profile`. They felt that otherwise it distracted from the recipe focus (I did pass on that the full info.json is better for efficiency reasons) 
 
If your OK to do the changes I think we can still get this recipe to the TRC in two weeks. Once you've done the changes Ill ask the editors to comment on it through slack rather than a meeting.

triplingual:
Can do, @glenrobson

triplingual:
✅ "could the title be changed to something like "Navigating in chronological order" so as not to have the property name in the title"
✅ "Could you cut down info.json of the image to just the id, type and profile."

_includes/links.md
```diff
@@ -24,6 +24,7 @@
 
 [0068]: {{site.baseurl}}/recipe/0068-newspaper/ "A newspaper"
 [0117]: {{site.baseurl}}/recipe/0117-add-image-thumbnail/ "Manifest Thumbnail"
+[0230]: {{site.baseurl}}/recipe/0230-navdate/ "Chronology with navDate"
```
glenrobson, line 27: Could you update this with the new recipe title?

triplingual: Yes indeed. While I'm at it, is that a good section for this recipe? My apologies for not raising it at the last authors' call.

glenrobson:
There looks to be some issues with the view of the json in the preview: https://preview.iiif.io/cookbook/0230-navdate/recipe/0230-navdate/

Do you know whats up? I'm breaking for dinner now but if you want me to take a look let me know. Do you need me to upload another image to fixtures?

triplingual:
And yes, @glenrobson, if you would add [https://www.loc.gov/resource/g3842c.ct008614/](https://www.loc.gov/resource/g3842c.ct008614/) to Fixtures

glenrobson:
Here is the fixture: https://fixtures.iiif.io/info.html?file=/images/loc/chesapeake_map/87691274-1986.jpg

recipe/0230-navdate/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Restrictions
+
+None known.
+
+## Example
+
+This recipe presents an imaginary Collection containing 2 instances from the run of a particular national park map held at the United States' Library of Congress. If you want to expand this example locally, you may find other maps in this series, at the [Library of Congress website](https://www.loc.gov/maps/?q=Chesapeake+and+Ohio+Canal&fa=contributor%3Aunited+states.+national+park+service&st=list&c=100).
+
+{% include manifest_links.html viewers="UV, Mirador" manifest="navdate-collection.json" %}
```
glenrobson, line 31: Could you put the links below the collection header?

recipe/0230-navdate/index.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Collection
+
+{% include jsonviewer.html src="navdate-collection.json" config='data-line="19"' %}
+
+### 1986 Map
+
```
glenrobson, line 38: Could you include the manifest_links.html include here?

recipe/0230-navdate/index.md
```diff
@@ -0,0 +1,56 @@
...
+{% include jsonviewer.html src="navdate_map_2-manifest.json" config='data-line="10"' %}
+
+### 1987 Map
+
```
glenrobson, line 42: and here if thats OK.

_includes/links.md
glenrobson: Maybe leave it where it is for now but I think we need to look at the headings now we have so many done in the first section.

recipe/0230-navdate/index.md
triplingual: I've placed a `manifest_links.html` include under each header, with the `manifest` attribute value of the adjacent Manifest, if that's how you meant this.

recipe/0230-navdate/index.md
triplingual: Added

recipe/0230-navdate/index.md
triplingual: Added

glenrobson:
Perfect thanks Trip!

glenrobson:
Approved with no changes requested by TRC.

PR #233, opened 2021-03-15
Supplementing motivation for the caption file.

mcwhitaker:


recipe/0219-using-caption-file/manifest.json
```diff
@@ -27,18 +27,25 @@
...
+                    "type": "Text",
+                    "format": "text/vtt",
+                    "label": {"en": [ "Captions in WebVTT format" ]},
+                    "language": "en"
+                  }
+                ]
               }
-            ]
+            ],
+            "target": "{{ id.path }}/canvas"
```
glenrobson, line 48: I think this line needs to be on line 46 before the `}` so its within the annotation.

recipe/0219-using-caption-file/index.md
```diff
@@ -27,8 +27,7 @@ Formats other than [WebVTT](https://w3c.github.io/webvtt/) (Web Video Text Track
 
 In this example we use a caption file in the WebVTT format, but one could just as well use a subtitle file in the [SRT](https://en.wikipedia.org/wiki/SubRip) (SubRip Text) or [TTML](https://w3c.github.io/ttml3/index.html) (Timed Text Markup Language) formats, or other text-based format used for the same purpose.
```
glenrobson, line 29: Could you add the link back to the manifest so it looks like this:

{% include manifest_links.html viewers="" manifest="manifest.json" %}

Issue #234, opened 2021-03-26
Identify content contributor with `provider`


triplingual:
## Identify content contributor with `provider`

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
You have a IIIF resource for which you would like to include information about one or more organizations or people that contributed to providing the content of the resource. You want to include a rich set of information for each contributor so clients can make visible this information in a similarly robust way.

PR #235, opened 2021-03-26
Recipe #0234—Identify content contributor with provider

triplingual:
This PR addresses issue #234, introducing the way to communicate rich content contributor information to a client for end viewer display.


glenrobson:
Comments from Editors

They are happy for this to go through to the TRC but had the following small edits:

## Use Case
* Change elastic to extensible in "The provider property allows you an elastic way"
 * Change: name and as much as a name - `label` and as much as a `label` in the same sentence as above. 
 * In the part which says "one or more `logo`s" the plural of logo looks a bit strange although we agreed it was correct implementation of the style guide. We wondered if we could avoid the plural by either saying "one or more `logo` properties or just say one logo as this recipe isn’t focused on the logo property. There was also some discussion on what clients would do with multiple logos....

## Example

 * Remove mirador as it doesn't currently implement provider. 

## Manifest 
 * Manifest ID has a space at the end 
 * Update label in seeAlso now we are using LOC metadata. (Editors agreed we didn't need to add MADS to the registry as this isn't the main point of the recipe)

I think we added most of above points in hypothes but the above are from the meeting notes.

triplingual:
Use Case
- [ ] Change elastic to extensible in "The provider property allows you an elastic way"
- [ ] Change: name and as much as a name - label and as much as a label in the same sentence as above.
- [ ] In the part which says "one or more logos" the plural of logo looks a bit strange although we agreed it was correct 
implementation of the style guide. We wondered if we could avoid the plural by either saying "one or more logo properties or just say one logo as this recipe isn’t focused on the logo property. There was also some discussion on what clients would do with multiple logos....

Example
- [ ] Remove mirador as it doesn't currently implement provider.
- [ ] Remove text with Mirador use directions

Manifest

- [ ] Manifest ID has a space at the end
- [ ] Update label in seeAlso as now we are using LOC metadata

regisrob:
Late feedback on recipe #234 (https://iiif.io/api/cookbook/recipe/0234-provider/): in the "Related Recipes" section, the link to "Linking to Structured Metadata" recipe is wrong (currently points to the navDate recipe)

PR #236, opened 2021-03-26
Image thumbnails on canvases

kirschbombe:


glenrobson:
Add two canvas to both examples

kirschbombe:
- The level 1+ and h/w comments have been resolved. 
- Per cookbook authors discussion, edited to mention CDN, but transformed to a note saying that discussion of pre-caching/fast access is out of scope for this implementation discussing.
- The cookbook authors also decided that a discussion of `extraFeatures` was out of scope and warrants another recipe
- Added a second canvas to video manifest to force thumbnails to display in viewers
- Will add a second canvas to the image manifest once Glen has the image up in fixtures

kirschbombe:
Second canvas added to image manifest example. Should be ready for TRC if the edits look OK.

Issue #79, opened 2019-01-08
Using Annotations for Timed Text


tomcrane:
## Using Annotations for Timed Text

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
You have WebVTT data for audio or video material. How do you transform that into web annotations?

bencomp:
I think I have a highly similar use case, or perhaps an extension. In my use case, the existing data are multi-layer annotations in the [ELAN Annotation Format (PDF)](http://www.mpi.nl/tools/elan/EAF_Annotation_Format.pdf). Annotation layers ('tiers') can be hierarchical, and annotations may be references to controlled vocabularies or simply textual.

stephenwf:
@bencomp Do you think that structure would fit into ranges with each range having annotations on them? Perhaps annotation collections could be used at that point under the ranges to denote layers.

```
{
  "id": "https://example.org/manifest",
  "type": "Manifest",
  "structures": [
      {
         "id": "https://example.org/manifest/r1",
         "type": "Range",
         ....
         "annotations": [
            {
               "id": "https://example.org/manifest/r1/annoPage1",
               "type": "AnnotationPage",
               "items": [ 
                   ... converted annotations ...
                ],
               "partOf": {
                   "id": "https://example.org/manifest/layer1",
                   "type": "AnnotationCollection"
                }
            }
            ...
         ]
      }
      ...
  ]
}

```

bencomp:
@stephenwf I hadn't thought of that, but yes, that might work. The EAF spec does relate annotations on child tier to (parent) annotations in the parent tier, so I'm not fully convinced yet :)

mcwhitaker:
The Cookbook Editors' suggestion is to broaden the scope of this recipe, mentioning in the Implementation notes that we are using WebVTT in the example but other formats could also be used.

stephenmcconnachie:
> The Cookbook Editors' suggestion is to broaden the scope of this recipe, mentioning in the Implementation notes that we are using WebVTT in the example but other formats could also be used.

That's great @mcwhitaker thank you. I have lots of subtitle files in various formats -SRT and TTML, and EBU-STL for that matter - and I also have plenty of WebVTT too. If it's useful, I could request permission to provide a real world a/v file with subtitle files in all formats I have.

elynema:
[...]

PR #239, opened 2021-04-30
Creating draft of recipe for dicsussion

Dananji:
Fixes issue #79


glenrobson:
Close as I think this is now in: https://github.com/IIIF/cookbook-recipes/pull/596 but happy to re-open if needed.

Issue #21, opened 2018-12-27
Tagging


tomcrane:
## Tagging

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
Examines the ways in which tagging annotations can be used in IIIF

benwbrum:
We are currently in the middle of work involving tagging annotations--users use the _Add Tag_ feature of Mirador 2 on their annotations, which are then presented in a IIIF v2/OA API and then consumed to create an index to annotations in a static site.  Once we're done with this, we might be able to translate the results into v3 for this recipe.

glenrobson:
Example from https://www.w3.org/TR/annotation-model/:

```
{
  "@context": "http://www.w3.org/ns/anno.jsonld",
  "id": "http://example.org/anno18",
  "type": "Annotation",
  "body": {
    "type": "SpecificResource",
    "purpose": "tagging",
    "source": "http://example.org/city1"
  },
  "target": {
    "id": "http://example.org/photo1",
    "type": "Image"
  }
}
```

glenrobson:
Example annotation in the spec: https://iiif.io/api/presentation/3.0/#56-annotation

PR #244, opened 2021-05-28
Recipe #0021 — Tagging

triplingual:
This PR addresses issue #21, introducing tagging as a form of annotation

### DIFF
--- a/recipe/0021-tagging/index.md
+++ b/recipe/0021-tagging/index.md
@@ -1,9 +1,9 @@
 ---
 title: Name of Recipe
-id: -1
+id: 21
 layout: recipe
-tags: [tbc]
-summary: "tbc"
+tags: [annotation]
+summary: "Tagging as a basic annotation"
 ---
 
 ## Use Case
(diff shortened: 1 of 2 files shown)


glenrobson:
Comments from the editors:

In the manifest:
 * `annotations` property should be in the Canvas rather than at the manifest level. 
 * Remove `summary`, `rights` and `requiredStatement` as they are not required.

Suggest we remove: “[Annotation in the context of a particular content resource][0023]” as related

triplingual:
>     * Remove `summary`, `rights` and `requiredStatement` as they are not required.

I believe these have been in other recipes using the Göttingen photo. It seems to me they are good practice that IIIF should encourage. But I'm OK with saving that for a discussion for a later time, when there's a read-through for standardization.

robcast:
@triplingual I just tested with Mirador and the annotation seems to work: https://projectmirador.org/embed/?iiif-content=https://preview.iiif.io/cookbook/0021-tagging/recipe/0021-tagging/manifest.json

I guess you can now include the automatic Mirador link and remove the sentence "No current viewers support this form of annotation..." :-)

triplingual:
Yow, @robcast, that's great! I tried it in the special annotations netlify, and it wasn't looking good, so this is particularly nice to know.

glenrobson:
Thanks Trip. I'll mark this as ready to go to TRC and create the TRC issue. 

Regarding "Remove `summary`, `rights` and `requiredStatement` as they are not required" thanks for removing them and we should discuss what we should do going forward in the authors call. During the editors call we compared it to the simple image recipes which didn't have summary and rights you may well be right about us encouraging the use of them so we should document that in the style guide and go back and make the other examples consistent.

Issue #246, opened 2021-06-02
Is image/jpg a valid media type?


ksclarke:
https://github.com/IIIF/cookbook-recipes/blob/993b3bb47aa019a81bd36e2c036ef16013f79701/recipe/0118_multivalue/manifest.json#L52

Shouldn't the media type for JPEG be image/jpeg (not image/jpg)? https://datatracker.ietf.org/doc/html/rfc2046#section-4.2

PR #249, opened 2021-06-16
Fixing mimetype from image/jpg to image/jpeg

glenrobson:
Reported by @ksclarke in #246


triplingual:
The first commit here will undo #250 if merged after it, yes?

Issue #245, opened 2021-05-29
Question about image format on 0004


ksclarke:
C.f. https://github.com/IIIF/cookbook-recipes/blob/993b3bb47aa019a81bd36e2c036ef16013f79701/recipe/0004-canvas-size/manifest.json#L24

Is the format intentionally "image/jpg" when the ID ends with "png"? (i.e., https://fixtures.iiif.io/video/indiana/donizetti-elixir/act1-thumbnail.png)

ksclarke:
And a follow-up... if it was supposed to be a JPEG format, isn't the mime/media type "image/jpeg", not "image/jpg"? Cf. https://datatracker.ietf.org/doc/html/rfc2046#section-4.2

glenrobson:
You are right it should be `image/png` as the mime type. Ill submit a pull request now. Thanks for reporting.

PR #250, opened 2021-06-16
Fixing image type to be png

glenrobson:
Reported by @ksclarke in #245


triplingual:
Looks good, displays good.

PR #252, opened 2021-06-29
Changing Newspaper recipe to be rendering rather than seeAlso

glenrobson:
This has been updated after discussions on the AV transcription use cases. For the initial change I've gone for `rendering` as the ALTO is a alternative representation of the page. This feels a better fit for the common use case with Newspapers. seeAlso was in appropriate because it isn't metadata about the resource.  

Another option was to include it as an annotation with the supplementing motivation. I got half way through this then I thought what would a viewer do with this? Then I remembered the Mirador text overlay plugin (https://mirador-textoverlay.netlify.app/) which supports ALTO and/or annotations. It currently takes the ALTO from the seeAlso but in 3 should it take the ALTO from a supplementing annotation like:

```
"annotations": [
        {
          "id": "{{ id.path }}/newspaper_issue_2-anno_alto",
          "type": "Annotation",
          "motivation": "supplementing",
          "body": [
          {
              "id":"{{ id.path }/newspaper_issue_2-alto_p1.xml",
              "type": "Dataset",
              "format": "application/xml",
              "profile": "http://www.loc.gov/standards/alto/",
              "label": {
                  "en": [
                      "ALTO XML"
                  ]
              },
              "language": "en"
          }
          ],
          "target": "https://iiif.europeana.eu/presentation/9200355/BibliographicResource_3000096302516/canvas/p1"
        }
```

but the overlay plugin could just as easily take it from the rendering property so I've gone with that for now.


glenrobson:
The cookbook editors are happy for it to go through but would like us to have a look at the OCR results. When you open the annotations in Annona they are all off where they should be. It could be:

 * Bad OCR data (which we could regenerate)
 * A problem with the conversion between Alto and annotations (the alto seems to have a margin which might be affecting it and also the width/height in the ALTO don't match the Image)
 * A problem with Annona although this is least likely it would be useful to see why it isn't working in Mirador

Other small comments on the related recipes:
 * Maybe remove: Simplest Manifest - Image as it would be include with every recipe 
 * Maybe include multi vol example

glenrobson:
I managed to get the ALTO working in https://mirador-textoverlay.netlify.app/ (by using `@id` rather than `id` and moving it to seeAlso and the coordinators are also off so I think its an issue with the OCR:

Screenshot 2023-03-08 at 13 00 36

glenrobson:
I also found the problem with Mirador is the motivation. If you switch it to `commenting` then the annotations show.

triplingual:
> I also found the problem with Mirador is the motivation. If you switch it to `commenting` then the annotations show.

(For future readers: `commenting` is not the appropriate motivation, just the one that allows the OCR annotations to function in Mirador at this time.)

glenrobson:
So as discussed could you move the partOf to each annotation similar to this example:

https://github.com/IIIF/cookbook-recipes/pull/416

and 

https://iiif.io/api/cookbook/recipe/0306-linking-annotations-to-manifests/

triplingual:
My batch F&R seems to have allowed auto-validation, and my eyeballing suggests that the annotations are in the same places and showing visually the same way.

PR #254, opened 2021-07-01
Newspaper fix

glenrobson:
Adding context to the mention of the Newspaper recipe.


glenrobson:
Page: https://preview.iiif.io/cookbook/news_fix/recipe/0231-transcript-meta-recipe/

@mcwhitaker if you are OK with the above change could you merge it into your branch? I thought because the Newspaper recipe covers many different use cases that it would be better to be explicit which part of the Newspaper recipe we were pointing to in this recipe.

mcwhitaker (approved):

Issue #261, opened 2021-07-26
Commenting Using a Non-Rectangular Polygon


triplingual:
## Recipe Name  
Commenting Using a Non-Rectangular Polygon

Relates to [2.x fixture](https://github.com/IIIF/api/blob/main/invalid_content_to_save/annotations.md#64-non-rectangular-segments)

## Use case  
I want to highlight a non-rectangular Canvas region visually.

Issue #260, opened 2021-07-23
Simple Commenting Annotation


glenrobson:
## Recipe Name  
Commenting annotation on a canvas

Relates to previous fixture: https://iiif.io/api/presentation/2.1/example/fixtures/51/manifest.json

## Use case  
I want to show a comment on a canvas/page with fragment.

glenrobson:
We are considering closing this as its covered in other recipes. We will need to change the reference to this recipe in:

https://iiif.io/api/cookbook/recipe/0266-full-canvas-annotation/

We can change the link to point to the tagging one as an example of a fragment:

https://iiif.io/api/cookbook/recipe/0021-tagging/

Also add the following note:

"Please note that Mirador does not put a highlight on the image when you mouse over the Annotation in the sidebar because the Annotation targets the Canvas as a whole."

in the commenting full canvas annotation.

glenrobson:
We should now include a reference to the fragment selector recipe #20.

triplingual:
> Also add the following note:
> 
> "Please note that Mirador does not put a highlight on the image when you mouse over the Annotation in the sidebar because the Annotation targets the Canvas as a whole."
> 
> in the commenting full canvas annotation.

With some research, it looks like these are the viewer behaviors:
* Mirador: Highlight when using fragment
* Annona: Never highlight when whole canvas
* Glycerine: Highlight when using fragment and mousing over the image
* Theseus: Always highlight

Consequently, I'd like to propose changing the added text to
"Note that viewers may not add a visual indicator of the Annotation when it targets the Canvas as a whole, or may only add it under certain circumstances, such as when targeting with a fragment."

PR #262, opened 2021-07-26
Recipe #0261 Non-Rectangular Comment Polygon

triplingual:
This PR addresses #0261, commenting with a non-rectangular polygon visual annotation.


recipe/0261-non-rectangular-commenting/manifest.json
```diff
@@ -70,18 +70,23 @@
...
+          	"type": "TextualBody",
+          	"value": "Gansenliessel"
+          },
+          "target": {
           	"type": "SpecificResource",
-          	"full": {
-          		"id": "https://iiif.io/api/image/3.0/example/reference/918ecd18c2592080851777620de9bcb5-gottingen/full/max/0/default.jpg",
-          		"type": "image/jpeg"
+          	"source": {
+          		"id": "https://iiif.io/api/image/3.0/example/reference/918ecd18c2592080851777620de9bcb5-gottingen",
```
glenrobson, line 81: I think the source should be a canvas like so:
```
"source": {
          "id": "{{ id.path }}/canvas/p1",
          "type": "Canvas"
}
```
Although the spec has the following example:

```
{
     "type": "SpecificResource",
     "source": "https://example.org/iiif/book1/canvas/p3",
      "selector": {
            "type": "FragmentSelector",
            "value": "xywh=0,0,750,300"
      }
}
```
in https://iiif.io/api/presentation/3.0/#54-range
so you could just use:

```
"source": "{{ id.path }}/canvas/p1"
```

glenrobson:
I found the following note on the w3c annotations model on which tags should be used:

"Note
Implementers SHOULD use only commonly supported features of SVG that directly contribute to describing a region, rather than styling or transformation, in order to maximize interoperability between systems. It is NOT RECOMMENDED to include style information within the SVG element, nor Javascript, animation, text or other non-shape oriented information. Clients SHOULD ignore such information if present."

from: https://www.w3.org/TR/annotation-model/#svg-selector

triplingual:
Excellent! Thanks for the research and example.

recipe/0261-non-rectangular-commenting/manifest.json
triplingual: I'm getting a build error, `source sequence is illegal/malformed utf-8` when I have 
```
          "target": {
            "type": "SpecificResource",
            "source": "{{ id.path }}/canvas/p1"
          }
```

recipe/0261-non-rectangular-commenting/manifest.json
```diff
@@ -0,0 +1,96 @@
...
+                ]
+              },
+              "target": "{{ id.path }}/canvas/p1"
+            }
+          ]
+        }
+      ]
+    }
+  ],
+  "annotations": [
```
glenrobson, line 64: This needs to be in the canvas part.

glenrobson:
I've fixed the validator so this now deploys correctly.

triplingual:
A few notes on the build today:
+ Mirador seems to only accept `commenting` and `tagging` for the annotation, so I've dummied it up with `tagging` for now
+ The annotation placement is off. I don't know whether this is my SVG or Mirador, but I'll figure it out.
+ I dropped the bounding box because the SVG is the same size as the Canvas and the bounding box was seeming to have no effect.

glenrobson:
Sorry for not getting these to you sooner but here are the comments from the Recipe Editors:

 * "For a IIIF resource, you’d like to add"
    - I find this too informal. Replace "you'd" with "you would"
 * "you’d like to highlight"
    - same here
 * "described as Scalable Vector Graphic (SVG) markup"
    - described in the Scalable Vector Graphic (SVG) markup format.
 * "This recipe is not the place to discuss how to create this markup, but we can note that many software tools exist that can handle it. Among others, third-party graphics editors may have this capability, as well as a Mirador instance properly configured. In each case, the creating engine may generate valid SVG that does not display as you wish in a client."
   - not sure we need this.
 * "To ensure your SVG is valid markup,"
   - To ensure your SVG markup is valid
 * "It’s best here to quote from the IIIF Presentation 2.1 API, noting that references to particular manifest properties may have changed with 3.0:"
   -  It’s best here to refer to the IIIF Presentation 2.1 API, noting that references to particular manifest properties may have changed with 3.0."  (link to the spec and remove the quote)
 * "This approach should not be used to describe non-rotated rectangular regions"
    - This approach should not be used to describe non-rotated rectangular regions; these are best described etc, etc OR for these should follow the specifications for etc etc
 * "we are highlighting the statue on the top of a fountain in Göttingen"
    - we are highlighting a fountain with a statue on top in Göttingen
 * "The bounding box should not show up on the image in any way."
    - The client should not show the bounding box on the image

In the example it says "Because we want to put the polygon in a particular place on the Canvas, we need to use the fragment selector syntax to describe its bounding box" and I think this is incorrect. I think the fragment selector is there if SVG coordinates are related to the bounding box. e.g. the svg says its 10pixels from the top of the box so not relative to the full image. I think in this case the svg is related to the full image so the fragment selector shouldn't be there...

This is at the bottom and I don't think it needs to there: 

"The direct link to the fixture is a useful convenience."

Related recipe:
 * Could relate the simple commenting recipe.

triplingual:
@glenrobson No worries. I've addressed these, with some light pushback on the one around stripping text about creating SVGs. I still think it's important to tell novice readers that valid SVG may not render as they expect in a client, and to note this right after we say that IIIF doesn't have anything to do with SVG. 

Also, note, I adjusted the title. We never got to this facet of the recipe in the last call, but it needed improvement. Happy to tweak again after TRC sees it.

triplingual:
Oh, and I wasn't clear about what should be referred to for rectangular non-rotated polygons without taking more time to dive back into other recipes or the spec.

glenrobson:
Thanks for the supper quick turnaround! 

I think it is just to say in the restrictions something like:

This approach should not be used to describe non-rotated rectangular regions; these are best described  using the simple fragment recipe. 

Which is this one I think "Simple Commenting Annotation" recipe #260

triplingual:
Right -- THe second part sure, but didnt you say that the fragment also serves as something for a client that can't display the SVG? I didn't have anything in the text about that, but if it's the case, it should get added and the fragment retained.

Issue #266, opened 2021-08-19
Full Canvas Annotation


glenrobson:
## Recipe Name  
Full Canvas Annotation

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I want to annotate the full canvas without specifying a part of a canvas. Use commenting as the motivation. Links to #260

PR #268, opened 2021-08-24
0266 full canvas annotation

robcast:
This PR addresses #266.

### DIFF
--- a/recipe/0266-full-canvas-annotation/index.md
+++ b/recipe/0266-full-canvas-annotation/index.md
@@ -1,5 +1,5 @@
 ---
-title: Basic Annotation
+title: Simplest Annotation
 id: 266
 layout: recipe
 tags: [tbc]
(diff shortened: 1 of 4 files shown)


glenrobson:
Direct link here: https://preview.iiif.io/cookbook/0266-full-canvas-annotation/recipe/0266-full-canvas-annotation/

glenrobson:
Code and title to:

https://github.com/IIIF/cookbook-recipes/blob/master/_includes/links.md

and then add link to it in the root index.md

glenrobson:

Would it be possible to refocus the recipe to be a basic annotation rather than focusing on it being a full canvas annotation? This would mean:

 * Change title from: "Full Canvas Annotation" to "Basic annotation"
 * Change the use case to something like "I want to annotate a canvas with some text either a comment or a tag. In this recipe we annotating the full canvas but see [Simple Commenting Annotation][260] if you would like to target a specific part or region of the canvas. 

Other comments:
 * Remove “​​View in Universal Viewer | View in Tify | View in IIIF Curation Viewer” to only Mirador 
 * Remove Summary, rights and requiredStatement from the manifest as they are not required.
 * Related - as its basic add basic IIIF image, audio + video

robcast:
I integrated your suggestions @glenrobson. I changed the title to "Basic Annotation" but the directory name is still `0266-full-canvas-annotation`. Should that be changed too? (I don't know if that would break anything)

glenrobson:
Its up to you regarding changing the directory name. If you do you will need to also change it in the `links.md` but that should be the only place.

glenrobson:
I've checked the changes and they all look good apart from:

  * Related - as its basic could you add links to the basic IIIF image, basic audio and basic video recipes

If you get chance to make the change then great but I don't think its a major issue so Ill get this ready for the TRC meeting next week.

recipe/0266-full-canvas-annotation/index.md
```diff
@@ -0,0 +1,45 @@
...
+title: Basic Annotation
+id: 266
+layout: recipe
+tags: [tbc]
+summary: "tbc"
+---
+
+## Use Case
+
+I want to annotate a canvas with some text either as a comment or a tag. In this recipe we annotating the full canvas but see [Simple Commenting Annotation][260] if you would like to target a specific part or region of the canvas.
```
tpendragon, line 11: So small, but `we are annotating`

glenrobson:
Hi @robcast now the voting has finished and been approved you can make the changes requested by the TRC.

Issue #74, opened 2018-12-28
Video with captions in multiple languages


tomcrane:
## Video with captions in multiple languages

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I have more than one language available for subtitles, including the language of the video itself.

nfreire:
I volunteer to work on creating this recipe. We have an example from Europeana that can be used in the recipe.

glenrobson:
Thanks Nuno, I've assigned you to the ticket.

PR #272, opened 2021-10-08
#74 Video with captions in multiple languages

glenrobson:


glenrobson:
Notes:
 * Fix highlighting
 * Remove seeAlso
 * Remove Mirador from list of viewers if it doesn't support subtitles.

glenrobson:
I've updated this from master so the styles should be sorted.

Issue #258, opened 2021-07-20
Tagging Annotation with External Resource


triplingual:
## Recipe Name  
Tagging Annotation with External Resource

## Use case  
Where #21 gave the simplest tagging annotation, using a Textual Resource to label a feature in the image, this recipe would start to move into the preferable scenario, a tag that connects to an external web resource, Wikidata perhaps.

glenrobson:
Notes from the Cookbook editor call
 * See [Slack conversation](https://iiif.slack.com/archives/CPLTX1ZD0/p1630334013000900) between Rob and Niqui O’Neill
 * Solution for referenced vocabulary
 * Need conversations about use cases: 
    * What are purposes of people who are implementing linked data into their annotations?
    * This might lead to identifying a need for a new kind of body, e.g., a SpecificResource with a Label
    * Would also guide what gets written in the “Use Case” section of the recipe
    * Ask Niqui directly

Consider using Europeana pattern:
```
{
  "motivation": "oa:tagging",
  "body": [
    {
      "@type": "SpecificResource",
      "source": "http://sws.geonames.org/2988507"
    },
    {
      "@type": "TextualBody",
      "value": "Paris",
      "language": "en"
    }
  ],
  "target": "http://data.europeana.eu/item/09102/_UEDIN_214"
}
```
 * Include both SpecificResource and TextualBody, make sure recipe body has caveat about viewers not inferring any connection between multiple body parts

PR #276, opened 2021-10-21
Recipe #0258 Tagging Annotation with External Resource

triplingual:
This PR addresses #258, for creating a tagging annotation that refers to an external resource


glenrobson:
The cookbook editors are happy for this to go to the TRC next week but asked for the following changes:

 * Could you add code highlighting to the Manifest for the tagging annotation
 * Remove “Because each body is used for tagging, we have put that as the value of the motivation property on the Annotation rather than making use of the Specific Resource’s purpose property.” from the example description as we don’t think there is any mention of purpose else where and this confuses the recipe a bit. We don't think there is a need to mention purpose in this recipe as its not used.  
 * Could you add an explanation for the second body i.e. we want to add a label for the URI in the first body. Maybe in the implementation notes or example. It would be valid to have a single body with the URI but to make it easier for a viewer/client we have included a second body as a label.
 * Below the Manifest could you take out the text “The direct link to the fixture is a useful convenience.”

triplingual:
- [ ] Could you add code highlighting to the Manifest for the tagging annotation
- [ ] Remove “Because each body is used for tagging, we have put that as the value of the motivation property on the Annotation rather than making use of the Specific Resource’s purpose property.” from the example description as we don’t think there is any mention of purpose else where and this confuses the recipe a bit. We don't think there is a need to mention purpose in this recipe as its not used.
- [ ] Could you add an explanation for the second body i.e. we want to add a label for the URI in the first body. Maybe in the implementation notes or example. It would be valid to have a single body with the URI but to make it easier for a viewer/client we have included a second body as a label.
  - To address this one, I did a little editing instead. I already had
  > In addition to tagging the fountain with text giving its name, we are adding for this recipe an additional Annotation `body` that points to the Wikidata entry for the fountain.
  - and changed it to 
  > While the Annodation `body` that points to the Wikidata entry for the fountain is valid all alone, we've added an additional `body` to provide the viewer a natural language textual label to display to a person using that viewer.
  - It's not thrilling prose, but does it get at the thinking behind the comments?
- [ ] Below the Manifest could you take out the text “The direct link to the fixture is a useful convenience.”
  - 🥴

triplingual:
Connections:
08b0c51: https://github.com/IIIF/trc/issues/85#issuecomment-988988362, https://github.com/IIIF/trc/issues/85#issuecomment-987134174
10b4a68: https://github.com/IIIF/trc/issues/85#issuecomment-989006553
3b9af42: https://github.com/IIIF/trc/issues/85#issuecomment-987134174
537217b: https://github.com/IIIF/trc/issues/85#issuecomment-988985391

glenrobson:
Thanks @triplingual merging now.

For info when I merged the latest from master I had to add the `viewer` and `topic` front matter to get it to pass validation.

Issue #39, opened 2018-12-27
Image with CSS Rotation


tomcrane:
## Image with CSS Rotation

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I want to include an image on the canvas, but it's upside down (or needs other rotation)

tomcrane:
Equivalent to 2.1 fixture

https://iiif.io/api/presentation/2.1/example/fixtures/39/manifest.json

glenrobson:
Reference in iiif 2.1: https://iiif.io/api/presentation/2.1/#rotation

2 options one using css or second using image selector: https://iiif.io/api/annex/openannotation/

glenrobson:
Use this one if you want to rotate the image but the image service doesn't support rotation.

triplingual:
To be merged with #40 into a single cookbook recipe with 2 manifests

Issue #40, opened 2018-12-27
ImageApiSelector - rotating/cropping an image onto a canvas


tomcrane:
## Reusing an image service (ImageApiSelector)
(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I need to present rotated and cropped versions of a IIIF Image on canvases

tomcrane:
Equivalent to 2.1 fixture

https://iiif.io/api/presentation/2.1/example/fixtures/41/manifest.json

robcast:
For cropping see the 2.1 fixture

https://iiif.io/api/presentation/2.1/example/fixtures/36/manifest.json

triplingual:
@kirschbombe Need link to Annex

triplingual:
Make manifest do region and rotation, in part to distinguish from #39

triplingual:
Good fixture: https://iiif.io/api/image/3.0/example/reference/85a96c630f077e6ac6cb984f1b752bbf-0-21198-zz00022840-1-master/full/max/0/default.jpg

glenrobson:
focus on rotation and bring in css example.

PR #290, opened 2022-01-11
Recipe #0039 Image with Rotation

triplingual:
This PR addresses #39, for demonstrating image rotation. The issue specifies CSS, but the plan now is to demonstrate both CSS and Image Service approaches.﻿


triplingual:
Conceptually merging into #40, so closing

PR #295, opened 2022-01-31
Recipe #0040 Image Rotation Two Ways

triplingual:
This PR addresses #39 and #40, for demonstrating image rotation through use of and affordances of:
1. IIIF Image API service
2. CSS selectors and rules


glenrobson:
Cookbook editor comments:

 * Expand on restriction to say the image api server should support the rotation requested by the selector. Only level 2 requires support for 90 degree rotation but it is optional at all levels: https://iiif.io/api/image/3.0/compliance/#33-rotation. Note the css rotation method doesn't rely on the image service as the rotation would be done by the viewer so that could be a reason to use that method instead. 
 * In the meeting we got a bit lost on the example not only being a rotation but also a region of the image. The suggestion was to change the example so it only shows one image. 
    * This would require:   
    * Remove region from the target in the canvas `#xywh=155,-148,1642,2556`
    * Swap width and height in canvas (as its rotated by 90) 
 * The editors also suggested showing in the recipe an example of what it should look like before and after applying the selector as there is currrently no viewer support. 
 * When ready removing the link to viewers 

Glen: Regarding a fixture for this use case do you want me to split the image you are using:

https://fixtures.iiif.io/info.html?file=/images/UCLA/conoximent_individuals/0-21198-zz00022840-1-master.jpg

Into two seperate iiif images? You could then rotate the first page without selecting the region.

triplingual:
@glenrobson Splitting the image would be lovely, thanks. No point in distracting readers and every point in showing them what we would expect it to look like.

glenrobson:
I've split the image you were using into two images. The first one (to be rotated) is here:

https://fixtures.iiif.io/info.html?file=/images/UCLA/conoximent_individuals/0-21198-zz00022840-1-page1.jpg

the second page is here:

https://fixtures.iiif.io/info.html?file=/images/UCLA/conoximent_individuals/0-21198-zz00022840-1-page2.jpg

I did them both in case it was useful to have two images for the before and after.

glenrobson:
Hi Trip, I hope is OK i've quickly fixed the image links and also the href links (change `[` to `(` in the markdown) so its ready to be shared with the editors. I'm going to share it now.

glenrobson:
Looks like I need to fix the validator to let this one pass.

glenrobson:
Do you want to comment out this line:

https://github.com/IIIF/cookbook-recipes/blob/d0d6125bc5db98dd11aacfe460714182f7385b54/.github/workflows/preview.yml#L74

So that it deploys and we can send it to the cookbook editors? If they agree with the JSON I can change the validator.

glenrobson:
Comments from the Cookbook Editors:

 * This is clear and well-written. I did not notice any style concerns (capitalization, non-https links, backticks etc.)
 * The CSS style of the site made the use of <h3> headings inside the "Implementation Notes" <h2> not look different enough for me and I had to back up to understand. Perhaps I am just groggy, but going down to h4, or using some other subheading way of breaking it down may be more clear.
 * Correction:
`In the image service section of your manifest, you must use an @context field`

should be

`In the image service section of your manifest, you must use a @context field`

Two comments on it being useful to clarify where the coordinates system that you end up with. For example you use the unrotated coordinates in the canvas and the post rotated coordinates in the canvas. This means any annotations on the canvas would not be affected by the rotation. Original question:

 * I would personally come to this recipe to find out the best practices for annotating these types of resources. Should I unrotate all my annotations and make them on the unjangled image? Is the IIIF Service formal enough to assume that the Canvas is always transformed (since their is not any "level 0" compliance for PresiAPI)? I would assume that all the changes made in CSS or ApiSelector would be made before annotating, but it would be reassuring to see that here, maybe in the Implementation Notes.

Is the image correct for the second manifest (rotate by CSS)? It looks like it is currently:

https://iiif.io/api/image/3.0/example/reference/85a96c630f077e6ac6cb984f1b752bbf-0-21198-zz00022840-1-master/full/500,/0/default.jpg

But I think it should be the same as the first manifest.

triplingual:
Resolving [TRC issue](https://github.com/IIIF/trc/issues/90) above comment, and Hypothes.is comments (to the extent these are non-identical):
1. [1098256676](https://github.com/IIIF/trc/issues/90#issuecomment-1098256676)
    Chose to change "image servers supports" to "image server supports" rather than "image servers support" as in the comment, as the rest of the graf talks about server in the singular.
1. [1098263233](https://github.com/IIIF/trc/issues/90#issuecomment-1098263233)
    Changed recipe text to address 1st and 2nd bullet points, going beyond suggested text changes for clarity.
1. [1100228294](https://github.com/IIIF/trc/issues/90#issuecomment-1100228294)
    Switching the grafs is a good question. My relatively consistent pattern in my recipes has been to enumerate the parts of a Manifest pertinent to a recipe first. However, for IIIF functionality without significant support such as rotation, it does make sense to put the support discussion first. This is a cookbook-level question as time goes on.  
    The graf mentioning region rotation is important for a thorough discussion of rotation capacities of an image service, I feel. It's true that the reader could learn that by reading more on their own, but I think I would like to leave it in on the assumption that people don't read.
1. https://hypothes.is/a/HHImGrtGEey-eDNtUwgjPg
    Punctuation changed.
1. https://github.com/IIIF/cookbook-recipes/pull/295#issuecomment-1108810661
    Expanded section text to explain more

Issue #300, opened 2022-02-17
Extra space in 0117-add-image-thumbnail manifest ID


ksclarke:
## Recipe Name  

0117-add-image-thumbnail

https://iiif.io/api/cookbook/recipe/0117-add-image-thumbnail/

## Issue

The manifest ID in https://iiif.io/api/cookbook/recipe/0117-add-image-thumbnail/manifest.json has an additional space between the ID and the end quote. See below:

`"id": "https://iiif.io/api/cookbook/recipe/0117-add-image-thumbnail/manifest.json ",`

I see in the source file, the JSON is generated from `{{ id.url }}` but I don't know the processes well enough to know where that variable (with its extra space?) comes from (or I would have submitted a PR for that).

triplingual:
Issue #300! Congratulations @ksclarke!

PR #302, opened 2022-03-09
Fix extra space

glenrobson:
Closes https://github.com/IIIF/cookbook-recipes/issues/300


ksclarke:
Doh, sorry I missed that space. Could have submitted a PR after all.

Issue #31, opened 2018-12-27
Bound multi-volume work


tomcrane:
## Bound multi-volume work

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
Are we talking Sammelband (for which there is a later issue) or are we talking a single-volume _Lord of the Rings_?

kirschbombe:
I'm thinking a good example might be music scores that are often bound together, although I wouldn't call it a "multi-volume work", maybe something like "multiple works bound together"? "multiple works in a single binding"? The example would use a single manifest with ranges (as opposed to multiple manifests brought together with a collection)

PR #303, opened 2022-03-09
Recipe #31: Bound multi-volume work

kirschbombe:
Preview: https://preview.iiif.io/cookbook/0031-bound-multivolume/recipe/0031-bound-multivolume/


kirschbombe:
I reduced the example further to just the front matter and first 2 volumes, plus added in @cubap 's summary (minus the 3rd volume and back matter).

Issue #299, opened 2022-02-04
Addressing a Region


triplingual:
**Recipe description mooted but left for comment thread context purposes. New description [in comment below](https://github.com/IIIF/cookbook-recipes/issues/299#issuecomment-1077630335)**

## Recipe Name  
Image Rotation of a Region

## Use case  
You have a region of an image that you want to rotate by default in presenting it to viewers. For one example, your images are of codex spreads, and one side of the view has text content oriented perpendicularly to the other half.

Comments and/or refinements welcome.

glenrobson:
I think it could also just be that you want a region of a iiif image which isn't rotated. For example I want to create a manifest of Newspaper articles which are regions of various Newspaper pages.

triplingual:
Do you mean that this issue is not about rotation or are you saying that the use case includes both 1) regions that are undesirably rotated and ~should be rotated and 2) regions that are not undesirably rotated but that you want to rotate?

glenrobson:
I wonder if we should re-focus this recipe to be about regions without rotation as we are already doing rotation of a full canvas in #40. The focus and use case for this recipe could be on a region with a small mention in the implementation notes of rotation and linking to #40. 

What do you think?

glenrobson:
Discussion on call. Merge with css rotation ticket so we can compare and contrast

triplingual:
Further discussion elucidated that this recipe will become only for discussion addressing a region, without rotation. #39 and #40 will merge into #40 to discuss rotation as a use case and include both the image service approach and the CSS approach.

Title of this one to change accordingly.

triplingual:
**NEW DESCRIPTION**

## Recipe Name  
Addressing a Region

## Use case  
You have a region of an image that you want to present independently.

Comments and/or refinements welcome.

PR #307, opened 2022-03-22
Recipe #0299 Addressing a Region

triplingual:
This PR addresses #299, for putting a region of a IIIF resource on a Canvas by itself


glenrobson:
The editors had a look at this recipe and I'm afraid found it quite difficult to parse the implementation notes and wondered if it could be simplified to give less options and focus on the recipe its self. 

Specifically they came up with the following suggestions:

 * Turn on validation: https://github.com/IIIF/cookbook-recipes/blob/c7e929292b2c1fcadbcefa00fa15391b5d5ae39a/.github/workflows/preview.yml#L73 although this has now been fixed in master so you can just update. 
 * First paragraph proposes many options and its unclear on which one we are going for. Could we focus on the details of the recipe and then in the final two paragraphs of implementation notes mention the other two options? Options include:
    -  Second option use a straight IIIF Image link like in recipe 001
    - Third option using something like content state (I'm not sure this is mentioned)
    - Forth option use an annotation like tagging recipe 

 * Suggestions for the following paragraphs:
 
“Note, however (as discussed in[ recipe 4](https://preview.iiif.io/cookbook/0299-region/recipe/0004-canvas-size/)) that the Canvas is a dimensionless set of coordinates. If desired, the dimensions of the Canvas may be different numeric values than the region you are displaying, but must be the same proportions if you want the region displayed with an unchanged dimension ratio.” 

  * Can we simplify? Something like the canvas dimensions will match the region of the sub image if you want the sub image to fill the entire canvas. (I think mentioning the dimensionless ness of the canvas will just confuse)
   
“When transmitting coordinates and dimensions for the region, note that the region property of the selector takes only those values, comma-separated, rather than the values in addition to a parameter type as the FragmentSelector uses” 

   * Note the region parameter of the image selector matches the parameters to the image API https://iiif.io/api/image/3.0/#41-region for example `x,y,width,height` or `pct:x,y,w,h`
  
“In addition, when using an Image service to retrieve your intended region of a IIIF resource, the Annotation body containing the resource is constructed as a[ SpecificResource](https://www.w3.org/TR/annotation-model/#specific-resources) containing in turn a source with the resource’s details.”

   * In this recipe the canvas resource (source IIIF Image) has the dimensions of the full sized image but then the region is pulled out by using a selector of type ImageAPISelector to retrieve the region from the source image. 

I'm afraid as there are quite a few suggestions the editors would like to see it again before it goes to TRC.

triplingual:
Totally fine to have a try at the cx and get it to them again. Comprehensibility is a fairly low bar for recipes. 🙂

triplingual:
Validation addressed with 5dcf26b

glenrobson:
Cookbook editors are happy for this to go through with the following notes. They appreciated the work you have done and found it much clearer to read. 

 * Maybe one to discuss in the cookbook meeting tomorrow but the editors were wondering whether we should add a new heading for ‘Alternatives’ or `Other solutions` to group the alternative solutions. This may not be necessary for early recipes but now they are getting more complicated it may help to only have the solution to the recipe in implementation notes and call out alternatives in its own section. 
 * The editors didn't think Annona worked with this recipe.  
 * In the restriction section could you link to this recipe when talking about timecodes in target as one example: https://preview.iiif.io/cookbook/0299-region/recipe/0064-opera-one-canvas/

glenrobson:
Brian makes a good point that this isn't in the context but I think in the rotation recipe we decided to go ahead as the editors are looking at updating the version 3 context. 

Glen will flag this with the Editors as we now have 2 recipes that use ImageAPISelector and soon will be 3.

Issue #283, opened 2021-12-17
Dealing with Missing images


glenrobson:
## Recipe Name  
Missing images in sequence. 

## Use case  
I have  a manuscript which has an image missing. I want to included a canvas with some content to show the user that there should/could be an image there and also make sure the recto/verso paged functionality isn't thrown due to this missing image. 

See this example: https://github.com/IIIF/api/issues/33

regisrob:
Live examples in Mirador 2 (manuscripts with missing pages represented as empty canvases):
https://demos.biblissima.fr/florus/florus-dispersus/mirador/
https://demos.biblissima.fr/bbmn-1713/mirador/

kirschbombe:
Thanks for the example links, @regisrob! I used the Biblissima examples last year to model how we approach this here at UCLA :)

glenrobson:
We also tried this at the NLW: https://glenrobson.github.io/iiif/2021/06/04/NLW-Journals.html

glenrobson:
Just looking at @regisrob example it looks like they also have empty canvases (from https://demos.biblissima.fr/iiif/metadata/florus-dispersus/manifest.json):

```
{

    "@id": ["http://demos.biblissima-condorcet.fr/iiif/metadata/florus-dispersus/canvas/canvas-1"](http://demos.biblissima-condorcet.fr/iiif/metadata/florus-dispersus/canvas/canvas-1),
    "@type": "sc:Canvas",
    "label": "Lacune (32r)",
    "width": 860,
    "height": 1200

},
```

@regisrob how do you get the custom thumbnail to appear? Is that a local customisation of Mirador? 

Also this example:
```
{

    "@id": ["https://iiif.biblissima.fr/bbmn-1713-t1/canvas/c-1"](https://iiif.biblissima.fr/bbmn-1713-t1/canvas/c-1),
    "@type": "sc:Canvas",
    "label": "p.1",
    "width": 3464,
    "height": 4898,
    "metadata": [
        {
            "label": "Cote",
            "value": "Page non retrouvée"
        },
        {
            "label": "Titre",
            "value": "Catalogus quorumdam Manuscriptorum Bibliothecae Vaticanae"
        },
        {
            "label": "Notice",
[...]

PR #317, opened 2022-05-02
#283 Missing images in a sequence

kirschbombe:
Issue #283 
Preview: https://preview.iiif.io/cookbook/0283-missing-image/recipe/0283-missing-image/


glenrobson:
Should now be deploying as I've commented out the validation while we work out which way to go:

https://preview.iiif.io/cookbook/0283-missing-image/recipe/0283-missing-image/

glenrobson:
Cookbook Editors are happy for this to go ahead but would suggest the following changes:

 * ID is capitalised in implementation notes but should be lowercase as its a property

Suggest rewritting:
"ID, type, height, width, and items properties. The first two are required by the API for any Canvas, the dimensions tell the viewer to maintain a viewing space of the appropriate size, and the final property ensures the widely used IIIF validator will pass the Manifest."

To something like:

"id, type, height, width, and items properties. The items property is empty in this example. The spec doesn’t require the items property but the IIIF-C validator will fail to validate a canvas without an items property as it differentiates between canvas references found in ranges and canvas found in sequences by the presence of items.

Feel free to re-word the above but it was felt it was slightly clearer on the soft requirment for items in a canvas.

Issue #316, opened 2022-04-26
Add Mirador viewer to #234 `provider`


triplingual:
See https://iiif.slack.com/archives/C18FHBKAQ/p1650474219141929 and https://iiif.slack.com/archives/CPLTX1ZD0/p1650474927237929

(Not quite a correction per se, but it's the best label.)

PR #320, opened 2022-05-18
Add Mirador viewer to #234 `provider`

triplingual:
Mirador has partial support for `provider`. The implementation is limited enough that the cookbook authors should discuss to make sure it is sufficient to add Mirador as a viewer for the `provider` property.
Addresses #316


glenrobson:
And only viewable when you view a list of manifest. Not the side panel.

glenrobson:
OK to be include with note above.

PR #321, opened 2022-05-18
Enhance svg discussion in #261

### LINKED ISSUE #305: Limited SVG selector recipe
# Limited SVG selector  

(_Note this is currently a rough gathering of thoughts_)

## Use case  

As an implementor it would it be valuable to distinguish between full SVG support and SVGs without masks (or even-odd cutouts), i.e. one or more paths. At the moment I cannot confidently implement #262 without either implementing _all_ SVG selectors or somehow documenting the limitation - which feel better suited as a cookbook entry.

In #262 it notes:
>  Note that client parsing of SVG varies, and valid SVG may not display as you expect.

I think this can be explained further using the existing SVG implementations, distinguishing between paths with and without cutouts.
The implementation notes also mention two ways that the sizing and placement can be affected, but only shows one of these in the cookbook. I think this example should highlight how viewBox could be used to give an additional, more reliable hint of the bounds of the selector. 

### Non-torus paths

Paths can be easily drawn directly on top of `<canvas />` elements, where as implementing the full SVG specification on a `<canvas/>` would be impractical or too heavy - for how often it would be used.  (See how [Mirador](https://github.com/ProjectMirador/mirador/blob/8a939b78e247d28fb1fd797b9c2b3968fef81e77/src/lib/CanvasAnnotationDisplay.js#L46) processes SVGs). The alternative would be to rasterize the the SVG or overlay the SVG (See how [Annotorious](https://github.com/recogito/annotorious/blob/main/src/selectors/EmbeddedSVG.js#L113) processes SVGs). 

The simplest implementation of drawing non-styled SVGs is to draw all `<paths/>`. However the SVG specification also allows for cutouts either using `<masks/>` or `even-odd` rules, which even `<path />` elements are additive, and odd `<path />` elements are subtractive. This is a marked jump in complexity for an implementation. 

I would propose adding a cookbook entry specifically for supporting non-torus paths, allowing implementations to more easily pick this as their chosen support level and balancing rich annotation tools and technical complexity. It would also give authors of IIIF resources outside of these tools guidelines to follow to create compatible SVGs.

### View box
Having an opinion or guidance on `viewBox` would also be valuable, and offer a quick fallback - but also reducing the size of the DOM element itself.

I'm not sure if `viewBox` would be better as an addition to #262 or not.

If a viewer was to overlay the SVG on top of the image, the interaction on that SVG would be the entire image, and it may be non-trivial to offer interactivity. (Instead of svg -> onClick, you would have to do svg -> path -> onClick, which also runs into complications with even-odd/cutouts). [Annotorious](https://github.com/recogito/annotorious/blob/main/src/selectors/EmbeddedSVG.js#L113) as worked around this by finding the areas of the SVG and assumes even-odd.  

I'm also not sure if this is indeed a standalone recipe or a limitation in scope for #262.

triplingual:
Changes to make the SVG discussion in "Annotation with a Non-Rectangular Polygon" more robust for certain implementors but not overly complex for typical or beginner Manifest creators.

Will resolve #305


glenrobson:
The cookbook editors are happy for this to go through to the TRC but request the following changes:

* There is a hypothesis comment but would it be possible to limit the use of the word resource in the use cases? Suggested re-wording:

“For a IIIF resource, you would like to add a simple annotation where the area you would like to highlight is not a rectangle.”

* Could you remove the following paragraph:

“Community feedback on the original version of this recipe led to conversations and thence to the decision to expand its discussion of SVG. If you have comments on this or any recipe, speak up in the #cookbook channel on the[ IIIF Slack workspace](https://iiif.slack.com/).” 

This was for two reasons, the first is that new readers won't be able to view the previous version of the recipe so will not understand the reference. For the link to the IIIF Slack we suspect iiif community readers will know how to discuss recipes and mentioning slack and a particular channel may not age well if discussions move on to the #annotations channel or even to discourse.

triplingual:
- [ ] reword use case
- [ ] remove graf

triplingual:
I adjusted the suggested use case rewording to use more IIIF-y words and improve readability.

PR #330, opened 2022-07-14
Remove need to manually enter viewers

dnoneill:


glenrobson:
I'm afraid we need to manually enter the viewers both in the front matter and further down in the recipe. The reason for this is that some recipes have multiple manifests e.g.:

https://iiif.io/api/cookbook/recipe/0232-image-thumbnail-canvas/

so we need to only show the viewer which works with the particular manifest we are referencing. The header viewer would likely then indicate partial support.

dnoneill:
@glenrobson The PR allows for that. By default it looks for included viewers, if they are not defined, then it will look for the page viewers. Just a thought. Feel free to close.

glenrobson:
Ah OK then it would be a great addition. Can it cope with structures like this:

```
viewers:
 - id: Mirador
   support: partial
```
from https://github.com/IIIF/cookbook-recipes/blob/master/recipe/0033-choice/index.md or would we need a rule that if it only includes patrial support then you must specify the viewer in the include?

glenrobson:
We should update the README.md to explain this new functionality. Best practice to leave viewers out of import if it is full support. Include if any partial support. 

But group agrees to merge.

triplingual:
README.md lines 24–27 and 78–83, and the section starting on line 46 are the focus areas, I believe

glenrobson:
Closing in favour of: https://github.com/IIIF/cookbook-recipes/pull/398/

PR #335, opened 2022-07-22
0326 annotating image layer

giacomomarchioro:
Merge test for 0326 annotating image layer


recipe/0326-annotating-image-layer/manifest.json
```diff
@@ -0,0 +1,98 @@
...
+                              "type": "Annotation",
+                              "motivation": "tagging",
+                              "body": {
+                                "type": "TextualBody",
+                                "value": "A group of skulls.",
+                                "language": "en",
+                                "format": "text/plain"
+                              },
+                              "target": {
+                                "source":"https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/max/0/default.jpg#xywh=810,900,260,370",
```
glenrobson, line 81: Cookbook editors suggest changing this to:


https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/2000,1271/0/default.jpg

recipe/0326-annotating-image-layer/manifest.json
```diff
@@ -0,0 +1,98 @@
...
+                      "service": [
+                        {
+                          "id": "https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-natural",
+                          "type": "ImageService3",
+                          "profile": "level1"
+                        }
+                      ]
+                    },
+                    {
+                      "id": "https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/max/0/default.jpg",
```
glenrobson, line 48: Cookbook editors suggest changing this to:


https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/2000,1271/0/default.jpg

glenrobson:
There was some uncomfortable-ness about the annotation targeting:

```
https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/max/0/default.jpg
``` 

although they agreed that this is the correct identifier, the concern was the URL contains the `max` size parameter which makes the annotation quite fragile. If `max` was to change dimensions either due to a larger image being digitised or a policy decision to allow larger image downloads then the annotation would break. 

The suggestion by the editors group was to use a URL like this in the target of the annotation and the image annotation id (which I've annotated in the pull request above):

https://iiif.io/api/image/3.0/example/reference/421e65be2ce95439b3ad6ef1f2ab87a9-dee-xray/full/2000,1271/0/default.jpg

The cookbook editors noted that if you are annotating someone else's manifest then you don't have control over this image ID and if it has `max` in it then you will just have to live with it but for the purposes of the recipe linking to the width and height of the image will make it more resilient to changes in max. 

We wondered if a discussion of this issue could be added to the restrictions part of the recipe.

A couple of other small notes:
 * It currently doesn't pass validation (may be a problem with the validator and I've just brought in the changes from master to turn validation back on)
 * Could you change the case of X-Ray to X-ray in the image label.  

Otherwise the cookbook editors were happy for this to go to the TRC. Let me know if you have any questions on the comments.

cubap:
> 

This is a strong enough concern that as a developer who works mostly with remote resources, I would not follow this pattern. It may be worth recommending using the `pct:` style for the `xywh=` FragmentSelector, which will not handle aspect ratio changes, but these are much less likely than scaled resolution changes.

giacomomarchioro:
> > 
> 
> This is a strong enough concern that as a developer who works mostly with remote resources, I would not follow this pattern. It may be worth recommending using the `pct:` style for the `xywh=` FragmentSelector, which will not handle aspect ratio changes, but these are much less likely than scaled resolution changes.

@cubap I also use pct: usually for avoiding this kind of issue maybe we can suggest to combine the two approaches.

recipe/0326-annotating-image-layer/manifest.json
giacomomarchioro: Is this supposed to be without the fragment selector (xywh=810,900,260,370)?

glenrobson: No sorry I didn't copy across the fragment selector but that should be in addition to the URI I put above.

glenrobson:
I've fixed the validator so it now accepts `scope` in an annotation.

glenrobson:
The cookbook editors reviewed the recipe and are happy for it to go back to the TRC with no changes.

They appreciated the simplification of the recipe and felt the image api selector was a better 'iiify' way of achieving the use case.

PR #331, opened 2022-07-19
Annona integration

dnoneill:

PR #336, opened 2022-07-22
Annona integration

glenrobson:
Moving https://github.com/IIIF/cookbook-recipes/pull/331 to the IIIF repo so it builds a preview


glenrobson:
Automatically show image if its composite: https://preview.iiif.io/cookbook/annona-integration/recipe/0036-composition-from-multiple-images/

Translate label  on title

dnoneill:
@glenrobson Changes integrated

PR #340, opened 2022-07-27
Removed redundant array around annotation bodies

stephenwf:
Changed to match other Annotations with single body annotations in the cookbook.


glenrobson:
Although not invalid to have an array we agree for consistency reasons we are happy to merge.

Issue #19, opened 2018-12-27
HTML in Annotations


tomcrane:
## Comments

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
Example of HTML in an annotation - needs a specific use case (annotation to link to / embed a cat gif/video?)

tomcrane:
Equivalent to 2.1 fixtures

https://iiif.io/api/presentation/2.1/example/fixtures/51/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/52/manifest.json
https://iiif.io/api/presentation/2.1/example/fixtures/54/manifest.json

glenrobson:
First one handled by https://iiif.io/api/cookbook/recipe/0266-full-canvas-annotation/

Second one handled by: https://iiif.io/api/cookbook/recipe/0021-tagging/

HTML in an annotation could be covered by #105 but maybe a basic HTML annotation would be helpful.

robcast:
Notes:
- Prezi3 talks about allowed HTML "in the `summary` property and the `value` property in the `metadata` and `requiredStatement`" https://iiif.io/api/presentation/3.0/#45-html-markup-in-property-values
- WebAnno uses HTML in example 5 of https://www.w3.org/TR/annotation-model/#embedded-textual-body and has no mention of allowed HTML

PR #344, opened 2022-08-05
Recipe #19: HTML in Annotations.

robcast:
This PR addresses recipe #19 about HTML in Annotation bodies.


glenrobson:

They are happy for this recipe to go through but have requested a small change to the following section:

```
 * Your HTML should be well-formed XML and therefore should be wrapped in an element such as p or span.
 * To alert a consuming application that your content is HTML, the first character in your string should be “<” and the last character should be “>”.
 * For security reasons, clients may allow only a, b, br, i, img, p, small, span, sub, and sup tags, and only href attributes on the a tag, and src and alt on the img tag, and may remove any or all of those.
```

There was concern that this list of allowed HTML tags might change in future versions of the specification particularly in relation to <div> (https://github.com/IIIF/api/issues/2122)  and it may be worth adding a note that div might be supported in the future to help with the first point on well-formed XML. It does open up a wider question on whether we will need to edit recipes when 3.1 comes out or indicate this is a 3.0 only recipe. 

For the third bullet point could you change the sentence to something like:

"These are the tags that the client may support and others and others maybe stripped"

The reason for this is there are other tags which are not security issues like `<em>` or `<strong>` that are not included in this list and viewers may or may not remove them. 

 
Let me know if you have any questions about the comments.

recipe/0019-html-in-annotations/manifest.json
```diff
@@ -0,0 +1,66 @@
...
+          "items": [
+            {
+              "id": "{{ id.path }}{{ id.path }}/canvas-1/annopage-2/anno-1",
+              "type": "Annotation",
+              "motivation": "commenting",
+              "body": {
+                "type": "TextualBody",
+                "language": "de",
+                "format": "text/html",
+                "value": "<p>Göttinger Marktplatz mit <a href='https://de.wikipedia.org/wiki/G%C3%A4nseliesel-Brunnen_(G%C3%B6ttingen)'>Gänseliesel Brunnen</a>"
```
stephenwf, line 57:
```suggestion
                "value": "<p>Göttinger Marktplatz mit <a href='https://de.wikipedia.org/wiki/G%C3%A4nseliesel-Brunnen_(G%C3%B6ttingen)'>Gänseliesel Brunnen</a></p>"
```

robcast: Oops 🤦

PR #358, opened 2022-09-23
Update id values for  Displaying Multiple Values with Language Maps 0118 recipe.

mathewjordan:
## What does this update?

While testing for recipe support I came across an error in the [Displaying Multiple Values with Language Maps (label, summary, metadata, requiredStatement)](https://iiif.io/api/cookbook/recipe/0118_multivalue/) recipe. Functionality in Clover using `@iiif/vault` requires `id` values to match.

This updates the Manifest `id` and the referenced `id` pattern within the `items` array to use jekyll variables `{{ id.url }}` and `{{ id.path }}` respectively. These previously were pointing to *https://example.org/iiif/text-language/* .

### DIFF
--- a/recipe/0118_multivalue/manifest.json
+++ b/recipe/0118_multivalue/manifest.json
@@ -1,6 +1,6 @@
 {
     "@context": "http://iiif.io/api/presentation/3/context.json",
-    "id": "https://example.org/iiif/text-language/manifest",
+    "id": "https://iiif.io/api/cookbook/recipe/0118_multivalue/manifest.json",
     "type": "Manifest",
     "label": {
       "fr": [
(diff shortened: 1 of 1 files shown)


recipe/0118_multivalue/manifest.json
```diff
@@ -1,6 +1,6 @@
 {
     "@context": "http://iiif.io/api/presentation/3/context.json",
-    "id": "https://example.org/iiif/text-language/manifest",
+    "id": "https://iiif.io/api/cookbook/recipe/0118_multivalue/manifest.json",
```
glenrobson, line 3: Could you use "{{ id.url }}" as this should be filled with the correct location for both preview and iiif.io

recipe/0118_multivalue/manifest.json
```diff
@@ -33,25 +33,25 @@
     },
     "items": [
       {
-        "id": "https://example.org/iiif/text-language/canvas1",
+        "id": "https://iiif.io/api/cookbook/recipe/0118_multivalue/canvas1",
```
glenrobson, line 36: Could you use "{{ id.path }}/canvas/1" again so it builds on both preview and iiif.io

glenrobson:
Hi Mat,

Thanks for catching this and happy to merge in principle but could you use the jekyll variables so that it works both on preview and iiif.io. I added comments above and you can see this variables in this example:

https://github.com/IIIF/cookbook-recipes/blob/master/recipe/0009-book-1/manifest.json

Thanks

Glen

glenrobson:
Thanks for making those changes they look good to me.

mathewjordan:
> Thanks for making those changes they look good to me.

Thanks Glen for pointing me toward those.

Issue #47, opened 2018-12-28
Homepage property: linking to a web page


tomcrane:
## Homepage

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  
I want to provide a link to a web page about the object.

glenrobson:
Homepage on the manifest.

Homepage in provider is covered by: https://iiif.io/api/cookbook/recipe/0234-provider/

glenrobson:
Relevant part of spec: https://iiif.io/api/presentation/3.0/#homepage

PR #362, opened 2022-09-30
Recipe #0047 Linking to Web Page of an Object

mathewjordan:
This PR addresses https://github.com/IIIF/cookbook-recipes/issues/47, setting the `homepage` property of a IIIF Manifest.

### DIFF
--- a/_includes/links.md
+++ b/_includes/links.md
@@ -30,6 +30,7 @@
 
 [0040]: {{ site.cookbook_url | absolute_url }}/recipe/0040-image-rotation-service/ "Image Rotation Two Ways"
 
+[0047]: {{ site.cookbook_url | absolute_url }}/recipe/0047-homepage/ "Linking to Web Page of an Object"
 [0053]: {{ site.cookbook_url | absolute_url }}/recipe/0053-seeAlso/ "Linking to Structured Metadata"
 [0046]: {{ site.cookbook_url | absolute_url }}/recipe/0046-rendering/ "Providing Alternative Representations"
 [0064]: {{ site.cookbook_url | absolute_url }}/recipe/0064-opera-one-canvas/ "Table of Contents for Multiple A/V Files on a Single Canvas"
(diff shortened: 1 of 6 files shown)


glenrobson:
Comments from cookbook editors:

In the first sentence:
“You have a IIIF Manifest representing an object that is also represented by one or more web pages” 

could you make it more explicit that this is the home page of the digital item. Could you also update the label in the manifest so that it mentions the home page of the digitial item as catalogue page may be more suitable in the seeAlso.

Could you change the example to use a image from https://fixtures.iiif.io rather than one hosted by Yale? Feel free to choose another one but this has a homepage at the Getty: https://fixtures.iiif.io/info.html?file=/images/Getty/laocoon.jpg. 

Other things:
 * Highlighting off in manifest. Should go to line 24. 
 * No need for thumbnail in Manifest. 
 * Add link to [metadata on any resource](https://preview.iiif.io/cookbook/0047-homepage/recipe/0029-metadata-anywhere/) in related recipes
 
Cookbook editors agreed with the above changes it was OK to go through to the TRC.

mathewjordan:
> Comments from cookbook editors:
> 
> In the first sentence: “You have a IIIF Manifest representing an object that is also represented by one or more web pages”
> 
> could you make it more explicit that this is the home page of the digital item. Could you also update the label in the manifest so that it mentions the home page of the digitial item as catalogue page may be more suitable in the seeAlso.
> 
> Could you change the example to use a image from https://fixtures.iiif.io rather than one hosted by Yale? Feel free to choose another one but this has a homepage at the Getty: https://fixtures.iiif.io/info.html?file=/images/Getty/laocoon.jpg.
> 
> Other things:
> 
>     * Highlighting off in manifest. Should go to line 24.
> 
>     * No need for thumbnail in Manifest.
> 
>     * Add link to [metadata on any resource](https://preview.iiif.io/cookbook/0047-homepage/recipe/0029-metadata-anywhere/) in related recipes
> 
> 
> Cookbook editors agreed with the above changes it was OK to go through to the TRC.

Thank you @glenrobson. The feedback has been applied and the Manifest now represents the Getty object you mentioned.

PR #383, opened 2023-01-31
Remove non-recipe link to Auth example

triplingual:
The one example we have in the Auth section links out not to a recipe but to an 
independent website. This pull would remove the link but leave the text.


glenrobson:
Change text to "basic auth example".

PR #388, opened 2023-02-03
0306 linking annotations to manifests

glenrobson:


glenrobson:
Cookbook editors are happy for this recipe to go through but note the following (which I think you are working on already):

 * Remove Annona from the View annotation page until it supports only passing the annotation URL and not the Manifest. Retain Annona on the Manifest viewers part
 * Fix the JSON-LD link so it shows the annotation list.  
 
The Annotation Page also fails validation at the moment:

Annotation list fails validation: https://presentation-validator.iiif.io/ at the moment but we think this is a validator issue and I will look into it.

glenrobson:
I've fixed the validator so the Annotation Page now validates correctly. Are you OK to make the other changes before the end of the day tomorrow?

Thanks

Glen

giacomomarchioro:
Hi @glenrobson , sorry for keeping you waiting I have removed Annona but I think that the second issue (Fix the JSON-LD link so it shows the annotation list.) should be already fixed, let me know if I am wrong so I can fix it in the same commit, otherwise I'll commit the patch for removing Annona.

glenrobson:
Yes, thank you the link to the JSON-LD looks good to me.

glenrobson:
Ready to merge.

PR #390, opened 2023-02-08
Add Annona to 299-region

dnoneill:


glenrobson:
Thanks Niqui this looks great! We have a cookbook meeting tomorrow and I will see if we can merge that then. Its great to have this recipe implemented in a viewer!

dnoneill:
@glenrobson Of course. Already had the functionality built in as a setting so it was pretty simple to integrate. The cookbooks have been great as a developers to have things to test against.

PR #394, opened 2023-02-14
fix annona view

dnoneill:


hadro:
This looks like it works to me, but I will leave it to @glenrobson to merge in as soon as he has a minute to look...

glenrobson:
I've tested all of the Annona links and they work with this change so I'm going to merge.

PR #413, opened 2023-05-19
0135 annotating point in canvas

glenrobson:


recipe/0135-annotating-point-in-canvas/manifest.json
```diff
@@ -0,0 +1,90 @@
+{
+  "@context": ["http://iiif.io/api/presentation/3/context.json",
```
glenrobson, line 2: I think this is causing the build error. It shouldn't have a , at the end.

kirschbombe:
Also, we will likely add the Selectors context to the Presi context, so there should be no need to import in a Manifest.

glenrobson:
The cookbook editors are happy for this to go to the TRC but request the following changes:

1. In the last paragraph of the Use case could you change "Viewer might consider" to "Viewers might consider" 
2. The code highlighting in the example looks a couple of lines to long.
3. Can you remove the full stop from the title.  So it reads "Annotating a specific point of an image"
4. In the use case can you change the tense to you rather than I so it would look like:

"You want to annotate a coordinate of the image rather than a region. You have a map with some locations that can be represented better by a single coordinate. You want to add some auxiliary data relative to a specific point of the image (e.g. some spectroscopic analysis made on a manuscript)."

This is to be consistent with other use cases in other recipes.

recipe/0135-annotating-point-in-canvas/index.md
```diff
@@ -0,0 +1,37 @@
...
+Viewer might consider implementing scale-independent point markers so that they are visible at every level of zoom.
+
+## Example
+
+This example uses a leaflet with a map and a guide supplied by the Library of Congress Geography and Map Division, it shows how we can annotate some locations expressed in the map.
+
+{% include manifest_links.html viewers="" manifest="manifest.json" %}
+
+{% include jsonviewer.html src="manifest.json" config='data-line="75-83"' %}
+
```
glenrobson, line 30: I'm not sure what is going on but it looks like this commit wasn't applied for some reason: https://github.com/giacomomarchioro/cookbook-recipes/commit/2287c460e439464939e13eb224ac369be25ba24f

glenrobson:
Looks like the changes have stuck now.

PR #426, opened 2023-08-16
Add UV to recipe 0005, basic image service example

hadro:


glenrobson:
Currently we use UV3 which doesn't support this recipe. We should look at moving to UV4 and evaluating the recipes it supports.

hadro:
That makes sense, thanks for taking a look

PR #433, opened 2023-09-29
0022 linking with a hotspot

glenrobson:


glenrobson:
Sorry for not reporting this earlier but the Editors are happy for this to go through to the TRC with the following small change:

 * Could you change http://example/ to https://example.com/ in the example external resource in the implementation notes?
 
 Ill send out the TRC agenda an issues on Monday but are you OK to make the meeting on Wednesday the 15th at 18:00 CT?
 
 Thanks
 
 Glen

giacomomarchioro:
No problem, just to be sure I have understood well, so instead of  `http://example.org/` which is the current version I'll write `https://example.com/`.

At 18:00 CET time/Rome, unfortunately, I am busy I am free until 17:30 CET time.  But feel free to present the recipe in my place.

See you soon, 
Giacomo

glenrobson:
I think this has some changes requested from https://github.com/IIIF/trc/issues/122

PR #439, opened 2023-10-18
Fix ImageService3 in 0047-homepage

ksclarke:
Looks like the service's type and profile are mixed? The link has the v3 version of the profile. I used the newer `id` and `type` keys too since this is a newer service.


glenrobson:
Thanks @ksclarke good catch! Merging now.

Issue #117, opened 2019-11-05
Add image thumbnail of resource


azaroth42:
## Image thumbnail of Resource

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  

I want to be able to display a thumbnail image for a resource, be that a canvas, a range, an audio file or a video, such that it can be clicked to activate that particular resource.

triplingual:
Does it matter whether this is declarative or via a service? Or should the recipe show both?

kirschbombe:
I vote for showing both :)

glenrobson:
We should change this one to be a manifest thumbnail and create new issue for canvas thumbnail.

PR #456, opened 2024-01-29
Update example text in #0117 for accuracy

### LINKED ISSUE #381: Old recipe's limitations about viewers
in #117, Mirador was the only viewer to support the use case when the recipe was written, and text was included that articulated that situation. This should get edited to reflect the addition of Clover as a supporting viewer.
Offending text:
> In Mirador, the sole viewer that uses it as of this writing, the Manifest thumbnail only displays when the site visitor uses “Add a resource” to change the loaded or active Manifests.

triplingual:
Closes #381


triplingual:
The primary impetus for this PR is 5f6f61c. The others are just gravy.

PR #469, opened 2024-03-21
Fix format to text/plain

### LINKED ISSUE #453: Format in recipe 0017 should be text/plain(?)
I believe the media type in recipe 0017's rendering format should be `text/plain` not `text/txt`.

https://github.com/IIIF/cookbook-recipes/blob/a89c866ac9d989d8a78e6de59bfc9932ef2ecc51/recipe/0017-transcription-av/manifest.json#L48

Somewhat related: @glenrobson mentioned in #450 that perhaps a test should be created for media types in cookbook recipes.

ksclarke:
* Update the format (media type) in 0017 from "text/txt" to "text/plain"
* Remove a couple of stray trailing EOL spaces

Fixes #453


glenrobson:
Thanks @ksclarke!

PR #471, opened 2024-03-24
Remove the hashes from the media-fragment values in the selectors.

### LINKED ISSUE #468: 0068 Newspaper's FragmentSelector(s) contain the hash delimiter
## Recipe info:

 * Name: 0068-newspaper
 * link: https://github.com/IIIF/cookbook-recipes/blob/master/recipe/0068-newspaper

## Correction request  

I'm not an expert on this so could be wrong, but am wondering if the FragmentSelectors in the newspaper recipes should contain the hash. For an example see https://github.com/IIIF/cookbook-recipes/blob/5cf4ddf2946528c8bde0507a2e8e75a8724b8c8b/recipe/0068-newspaper/newspaper_issue_1-anno_p1.json#L29

In it the FragmentSelector value is `"#xywh=0,376,399,53"`

If I look in the W3C Annotation Data Model documentation, though (https://www.w3.org/TR/annotation-model/#fragment-selector), the values don't contain the hash delimiters, but just the value itself (e.g. `xywh=50,50,640,480`).

I _think_ that the hash should not be included at the start of the FragmentSelector value(?) Is this incorrect?

ksclarke:
Fixes #468

### DIFF
--- a/recipe/0068-newspaper/newspaper_issue_1-anno_p1.json
+++ b/recipe/0068-newspaper/newspaper_issue_1-anno_p1.json
@@ -26,7 +26,7 @@
                 "selector": {
                     "type": "FragmentSelector",
                     "conformsTo": "http://www.w3.org/TR/media-frags/",
-                    "value": "#xywh=0,376,399,53"
+                    "value": "xywh=0,376,399,53"
                 }
             }
         },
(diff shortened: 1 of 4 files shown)


glenrobson:
Thanks again @ksclarke!

Issue #359, opened 2022-09-23
Adjust language about permitted markup elements in viewers, Recipe #7


triplingual:
"For security reasons, clients allow only" ==> "For security reasons, clients are only expected to allow only"

Helps with clarity, came up a bit in discussing #19

glenrobson:
Fixed in #477

PR #477, opened 2024-04-02
Clarify language about permitted markup elements in #7

triplingual:
Slight deviation from issue #359 text. Deleted one 'only' to avoid repetition


triplingual:
Closes #359

Issue #441, opened 2023-11-09
Update sentence on Mirador support in recipe 0234-provider


regisrob:
Mirador 3 now supports the display of provider `logo` (since https://github.com/ProjectMirador/mirador/pull/3657). 
But unfortunately the official demo is not up-to-date (it may be something that could be asked to maintainers at Stanford?).
When it is updated, [this sentence](https://github.com/IIIF/cookbook-recipes/blob/master/recipe/0234-provider/index.md?plain=1#L37) about Mirador support could be changed in the recipe. But it has no impact on the viewer matrix (support remains `PARTIAL`).

glenrobson:
Still the case that the deployed Mirador doesn't support this yet.

glenrobson:
Fixed in above merge.

PR #478, opened 2024-04-02
Add text noting support for `logo` to #234, `provider` recipe

triplingual:
Mirador still only partially supports `provider`, but it supports it a little more than before


triplingual:
Closes #441

Issue #474, opened 2024-04-02
Note that duration should be a floating-point number


triplingual:
## Recipe info:

 * Name: Simplest Manifest - Audio
 * link:  https://iiif.io/api/cookbook/recipe/0002-mvm-audio/

--

 * Name: Simplest Manifest - Video
 * link:  https://iiif.io/api/cookbook/recipe/0003-mvm-video/

## Correction request  
Need to add a comment that the duration needs to be a floating-point number. That fact is obscured in the recipe currently because the time is highly precise, but even if the duration were rounded to 1985 seconds, it would need to be written as 1985.0

PR #479, opened 2024-04-02
Add text about `duration` value datatype in #2 and #3

triplingual:
The value for a `duration` property must be a floating point number (per Prezi 3.0 API), and these recipes are the proper places to note that.


triplingual:
Closes #474

Issue #482, opened 2024-04-12
Clover now Supports 0046-rendering


markpbaggett:
## Recipe info:

 * Name: Providing Alternative Representations
 * link: https://iiif.io/api/cookbook/recipe/0046-rendering/

## Correction request  

Clover supports this as of [2.7.4](https://github.com/samvera-labs/clover-iiif/releases/tag/v2.7.4).

Here is an [example](https://samvera-labs.github.io/clover-iiif/docs/viewer/demo?iiif-content=https%3A%2F%2Fiiif.io%2Fapi%2Fcookbook%2Frecipe%2F0046-rendering%2Fmanifest.json).

PR #483, opened 2024-04-12
Address Issue-482.

markpbaggett:
This addresses [Issue-482](https://github.com/IIIF/cookbook-recipes/issues/482).


recipe/0046-rendering/index.md
glenrobson, line 1: You also need to add clover to the following line:
```
{% include manifest_links.html viewers="Mirador,Annona" manifest="manifest.json" %}
```
or remove the viewers attribute and it will take it from the header.

glenrobson:
OK to merge once Clover is added to the viewers link at the bottom.

PR #491, opened 2024-05-24
Add Glycerine Viewer

glenrobson:
From original [pull request](https://github.com/IIIF/cookbook-recipes/pull/490):

"Initially designed for the sophisticated annotation features of [Glycerine Workbench](https://glycerine.io/), [Glycerine Viewer](https://github.com/Systemik-Solutions/glycerine-viewer) has been released as an open source IIIF viewer to support IIIF manifests and compliant annotations built with other tools. Developed with a contemporary Vue 3 framework, Glycerine Viewer has an elegant interface, supports the IIIF Presentation API and provides the most comprehensive annotation feature set of any IIIF Viewer."

PR #496, opened 2024-05-27
Changes to recipe support of Glycerine Viewer

yangli0516:
Related to https://github.com/IIIF/cookbook-recipes/pull/491

- Have added the Glycerine Viewer into recipes 0068 and 0030
- Removed the "No viewers currently implement this tagging approach" from recipe 0258


glenrobson:
Great thank you @yangli0516!

Issue #489, opened 2024-05-17
Complex multimedia presentation 


triplingual:
## Recipe Name  
Complex multimedia presentation

## Use case  
_You want to combine video and image on a single canvas for digital storytelling._

triplingual:
🔥: https://tomcrane.github.io/fire/

PR #501, opened 2024-06-04
Added Theseus Viewer support

stephenwf:
https://theseus-viewer.netlify.app/

PR #507, opened 2024-06-07
Recipe #0489 Multimedia Canvas

triplingual:
This PR addresses #489, for making a Canvas that holds image and AV resources


triplingual:
No known viewer support. Viewers are currently listed to facilitate checking as the recipe progresses.

stephenwf:
@triplingual Theseus supports this recipe:

https://theseus-viewer.netlify.app/?iiif-content=https://preview.iiif.io/cookbook/0489-multimedia-canvas/recipe/0489-multimedia-canvas/manifest.json

There is also the Fire demo Tom created back in 2016, which feels relevant:
https://theseus-viewer.netlify.app/?iiif-content=https://tomcrane.github.io/fire/fire3.json

recipe/0489-multimedia-canvas/manifest.json
```diff
@@ -0,0 +1,57 @@
...
+                    "id": "https://iiif.io/api/image/3.0/example/reference/918ecd18c2592080851777620de9bcb5-gottingen",
+                    "profile": "level1",
+                    "type": "ImageService3"
+                    }
+                  ]
+                },
+                "target": "{{ id.path }}/canvas"
+              },
+              {
+                "id": "{{ id.path }}/annotation/p0002-video q",
```
stephenwf, line 39:
```suggestion
                "id": "{{ id.path }}/annotation/p0002-video",
```

recipe/0489-multimedia-canvas/manifest.json
```diff
@@ -0,0 +1,57 @@
...
+                "type": "Annotation",
+                "motivation": "painting",
+                "body": {
+				  "id": "https://fixtures.iiif.io/video/indiana/donizetti-elixir/vae0637_accessH264_low_act_1.mp4",
+				  "type": "Video",
+				  "format": "video/mp4",
+				  "height": 360,
+				  "width": 480,
+				  "duration": 3971.24
```
stephenwf, line 48: There is no selector on the resource, so because it's painted to: `t=27,57` I've assumed that the selector/cropping is `t=0,27`. Is that a correct assumption?

triplingual: Things are in-flight enough that it's a correct assumption given the current state of things, but it may become an incorrect one as I continue to edit. Right now, everything has been for placement only, to see whether I could get it syntactically valid and perhaps even working in a viewer. Now that you've shown me Theseus (thank you very much!) I can do something with the manifest and resources that might approach an authentic implementation of the use case. As part of that, I'll tighten up the syntax and make sure the property values are appropriate.

glenrobson:
I've had a look at the manifest and I think it is valid despite what the validator says.

I've tried it in the new validator version: https://github.com/IIIF/presentation-validator/pull/155 and it works fine. So I think for now turn off validtion to get it to build.

glenrobson:

Suggested title change:-  Teaching with a complex multimedia canvas

Editors thought the existing title "Multimedia Canvas" brought together a combination of image and av like the accompanying canvas rather than the use case of a complex multimedia canvas. 
 
## Implementation notes
 * Would it be possible to expand the implementation note section to talk more about how the annotations are targetd to a manifest using the idea of a timeline and mention selections. Also would be useful to note the order of the annotations in the manifest don’t match what you see in the viewer because of the target time stamps. 
 * Move third paragraph into restrictions “When using timing for showing…”
 * add a note to this pargraph to mention viewers may allow the user to control the timeline so you can view things after they’ve loaded. 

## Example
 * Remove () from last (These final instructions stay visible for approximately 2-1/2 minutes to avoid an immediate return to the initial instructions.)
 * Link to localhost in the Theseus Viewer temporary link (but note this can be replaced once #501 is merged)

The editors would like to look again before TRC so if it would be possible to have the changes done by Monday I can send it to them for comments.

Thanks

glenrobson:
Also note we tested enabling the validation and it seems it does now validate. So if you want to merge this pull request: https://github.com/IIIF/cookbook-recipes/pull/531 into this one it should re-enable the validation.

triplingual:
@glenrobson Could you say a bit more about this one?
> add a note to this pargraph to mention viewers may allow the user to control the timeline so you can view things after they’ve loaded
I'm reading it to say that control of the viewer is (may be) in the hands of the person viewing, as with Theseus, so there's no way in this example to prevent someone from jumping around the timeline or pausing the timeline. Another way it's not a drop-in example for teaching, just a general proof-of-concept.

triplingual:
- [ ] Change title to _Teaching with a complex multimedia canvas_
- [ ] Implementation Notes: Expand to talk more about how the annotations are targeted to a manifest using the idea of a timeline and mention selections
- [ ] Implementation Notes: Note that the order of the annotations in the manifest don’t match what you see in the viewer because of the target time stamps
- [ ] Implementation Notes: Move third paragraph into restrictions “When using timing for showing…”
- [ ] Restrictions: Add a note to this (moved, see above) paragraph to mention viewers may allow the user to control the timeline so you can view things after they’ve loaded.
- [ ] Example: Remove () from last (These final instructions stay visible for approximately 2-1/2 minutes to avoid an immediate return to the initial instructions.)
- [ ] Example: Replace Theseus Viewer temporary link

triplingual:
Not 100% sure I got the bit about timeline and selections, but it's a first pass. 
I also added a related recipe for a more basic example of time in `target` and a mention in the Example section about how the timer and picture are ordered in the Manifest to get the timer on top of the picture.

kirschbombe:
Hi, @triplingual - Re: the comment about "viewers may allow the user to control the timeline", that may have been a miscommunication. I think we were suggesting you add details to the implementation section to highlight the use of the time and fragment selectors to create the canvas timeline, which you've done :) Someone may have suggested something about user controls, but I don't remember. In any case, you do talk about this a bit in the restrictions, which is good.

kirschbombe:
For the title, if we don't want it to me specifically about teaching, perhaps "Time-based multimedia canvas" - or something that makes it clear that it's not just a canvas of time-based things (A/V), but that the canvas itself is time-based?

triplingual:
Yeah, the use case / title are a little tricky here. Bringing together multiple kinds of resources/Annotations also lends itself to digital storytelling, kiosks, and more. I'm reflexively accommodating to editing, so I changed the title, but it is worth some thought about what would signal to someone reading the cookbook matrix or TOC that there are a lot of directions to take with what you learn from this recipe.

kirschbombe:
@triplingual - Title suggestion from the editors " Rendering multiple media types on a time-based Canvas"

PR #525, opened 2024-08-02
First try for adding colour to the table

glenrobson:
Using the icons from the IIIF specifications.


glenrobson:
Icons look OK, maybe rename the image files so they aren't the same as the spec. 

Fix build and approve over slack.

glenrobson:
Add explination of Partial and text about  YES, NO or Partial at the top.

PR #529, opened 2024-08-09
0485 contentstate canvas region

glenrobson:
Creating pull request for a preview.


glenrobson:

## Use case
 * Suggested re-wording of first paragraph: "I want to share a link to highlight a detail in a specific region of an image to share a portion of a digitized manuscript text, allowing the user easy access to the rest of the manuscript."

## Implementation notes
 * Suggest removing “To implement this feature, we must create a request to an entry point of a viewer instance with the necessary information for the viewer to open the Manifest, retrieve the Canvas and set the viewport to the desired region.” as its covered in the recipe. 
 * Suggest moving the “[Image API](https://iiif.io/api/image/3.0/#41-region) allows users to share “ paragraph to the end of the implementation notes as it is good to note this possibillity it would be better to put it further down so as not to distract from the current use case. 
 * Could you move the text in example to implementation notes as it defines the process of creating the annotation which seems to fit better in implementation notes. 

## Restrictions

 * Useful to mention that content state doesn't define how the viewer should bring the regsion of the canvas to the users attention. It only mentions "This data structure can be used by clients to load the resource required, present a particular part of the resource to the user." https://iiif.io/api/content-state/1.0/#content-state. Viewers may set the viewport to the region or highlight the region with an annotation. 
 
## Example
 * Could you include an image in the example showing what part of the image would be highlighted so its possible to tell in a viewer if its showing the correct region. A mockup of what a possible viewer could do with this annotation might be useful. Something similar to: https://preview.iiif.io/cookbook/0489-multimedia-canvas/recipe/0139-geolocate-canvas-fragment/
 * Take the view links away. Clover is the only viewer that accepts the annotation but it doesn't zoom into the region so doesn't support this recipe yet. 
 
 The editors would like to see the updates before the TRC so if you could make the changes by the end of Monday I can feed it back to them for review. 
 
 Thanks

PR #530, opened 2024-08-13
Adding Content State encode

glenrobson:
Note this will only replace viewers that use the iiif-content parameter e.g. Mirador in the example


glenrobson:
For working example see the Mirador link in this recipe: https://preview.iiif.io/cookbook/add_content_state_links/recipe/0485-contentstate-canvas-region/

PR #541, opened 2024-09-27
Add multilingual link

glenrobson:
Adding a note to the bottom of the example in:

https://preview.iiif.io/cookbook/add_multilingual_link/recipe/0006-text-language/

to add the Curation viewer and clairfy what supporting the recipe means.

Also making Thesus and Curation partial for [Viewing direction and Its Effect on Navigation (viewingDirection)](https://preview.iiif.io/cookbook/add_multilingual_link/recipe/0010-book-2-viewing-direction/):

https://preview.iiif.io/cookbook/add_multilingual_link/recipe/matrix/

as both only support the right to left manifest not the top to bottom manifest.

### DIFF
--- a/recipe/0006-text-language/index.md
+++ b/recipe/0006-text-language/index.md
@@ -39,7 +39,11 @@ In the `metadata` and `requiredStatement` properties, the `label` strings for "C
 
 Finally, the `summary` property (lines 49-54) has values in both English and French.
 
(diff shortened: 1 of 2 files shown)


glenrobson:
Remove sentance:

> Note the viewers included in this recipe are for the specific languages in the example of this recipe

>  Universal Viewer, Mirador and Glycerine

add Oxford comma

PR #546, opened 2024-10-12
language should be an array

tomcrane:
This passes the validator so I was beginning to doubt myself, but just tripped up a parser of ours.

https://iiif.io/api/presentation/3.0/#language

### DIFF
--- a/recipe/0219-using-caption-file/manifest.json
+++ b/recipe/0219-using-caption-file/manifest.json
@@ -54,7 +54,7 @@
                     "Captions in WebVTT format"
                   ]
                 },
-                "language": "en"
+                "language": ["en"]
               },
               "target": "{{ id.path }}/canvas"
             }


stephenwf:
This ones tricky, because the IIIF specification does say it should be an array but the W3C model says it could be either.

The IIIF specification example _also_ has both - with the annotation being just a single value.
https://iiif.io/api/presentation/

Screenshot 2024-10-12 at 18 59 50
3.0/#b-example-manifest-response

W3C: https://www.w3.org/TR/annotation-model/#h-bodies-and-targets

glenrobson:
Will need to fix the validator to allow arrays and strings for this.

recipe/0219-using-caption-file/manifest.json
```diff
@@ -54,7 +54,7 @@
                     "Captions in WebVTT format"
                   ]
                 },
-                "language": "en"
+                "language": ["en"]
```
stephenwf, line 57: Because the annotation model (and the IIIF Specification itself) has single languages as string values in annotation bodies - I think this should too.

ahankinson:
As far as I remember when doing the v3.0 spec, it was decided? hoped? that by making values consistently an array, even if they were arrays of single values, it would help developer ergonomics since they could reasonably assume that everything was an array, rather than needing to constantly check the datatype for each field (which is what you needed to do in v2). It was noted at the time that (particularly for language maps) this diverged from the ways you could do it in the Web Annotation spec, but that IIIF could choose to be consistent.

I think we discussed this at the meeting in Toronto, specifically... I'm hoping I'm not misremembering.

PR #562, opened 2024-11-22
Recipe #560: Resources on a Timeline

### LINKED ISSUE #560: Displaying resources along a timeline independent of time-based resources
## Recipe Name  
Displaying resources variably over time

## Use case  
You want to make something like a slide show of resources to put in a display for non-interactive viewing.

triplingual:
Closes #560


glenrobson:
The cookbook editors are happy for this to go through but asked if you could look at the following:

# Implementation notes
 * “the viewer is expected to provide a means of displaying that Canvas...” could you be less directive as the cookbook generally encoruages viewers rather than directs what they should do. 
 * “Nothing in[ the spec](https://iiif.io/api/presentation/3) says anything explicitly about a viewer” - remove link to spec as it doesn't link to a specific section. 
 * There was discussion whether the t parameter should be in quotes rather than backticked 
 * “it can use a time fragment on its target value, in the form #t=” could you use media fragment instead of time fragment as this is the term in the mediafragments specification.   
 * “This parameter’s value can be a range, a starting time (to indicate the resource should display until the end of the duration), or an ending time (to indicate the resource should display from the beginning of the duration).” - For details see the media fragment spec when specifying one number. If we have to have it include the table from the media frags from spec. 
 * Add to the “Nothing in the spec paragraph” that other values are allowed in the t parameter but conventionally seconds are used. See slack discussion: https://iiif.slack.com/archives/C18FHBKAQ/p1663762334826879 and (IIIF api issue](https://github.com/IIIF/api/issues/2336)
 * Suggested change: “To get the Canvas to repeat the sequential items display, the behavior value used is repeat” “To get the annotation sequence to repeat the behavior value is set to “repeat”. 
 * Suggested change: “Within a Canvas’s items the parameterized target values tell viewers” - Within a Canvas’s annotations the parameterized target values tell viewers

# Example
The images in the manifest has different aspect ratios so you should include the note from the linked recipe: “The aspect ratio should be consistent between your source image and Canvas. Otherwise, you’ll see unpredictable stretching and/or distorting.”

triplingual:
Just to note that I will start tackling this 20 March.

triplingual:
Changes made per IIIF/trc/issues/132 comments

Issue #556, opened 2024-11-14
Add provides property to Using Caption and Subtitle Files in Multiple Languages with Video Content


elynema:
## Recipe info:

 * Name: Using Caption and Subtitle Files in Multiple Languages with Video Content
 * link: https://iiif.io/api/cookbook/recipe/0074-multiple-language-captions/

## Correction request  

The AV Annotations TSG would like to add the proposed [provides property](https://github.com/IIIF/api/issues/2308) to this recipe, allowing the IIIF client to distinguish between captions (original language, needed for accessibility) and subtitles (translations from original language). It is a very common use case for Europeana to provide both a caption and subtitle for audiovisual items.

In the recipe, the first file would be labeled with `provides='openCaptions'` and the second `provides='subtitles'`. Note that the term `subtitles` is not included in the original vocabulary for the `provides` property, and this is a demonstration of the use case for adding it.

Common usage (across IU and Europeana) when there are multiple languages for caption and subtitle files is to use multiple `supplementing` annotations, rather than a `choice` construct within a single annotation. Could we separate the `choice` construct into 2 annotations, each with its own `provides` property?

elynema:
If needed, the AV Annotations TSG could survey AV manifest providers to ask about their practices for multiple caption/subtitle files.

Issue #555, opened 2024-11-13
Add provides property to Using Caption and Subtitle Files with Video Content recipe


elynema:
## Recipe info:

 * Name: Using Caption and Subtitle Files with Video Content
 * link: https://iiif.io/api/cookbook/recipe/0219-using-caption-file/

## Correction request  

With the proposed new [provides property](https://github.com/IIIF/api/issues/2308), we'd like to add provides:openCaptions to the annotation containing the caption to demonstrate its use to the AV community.

In addition, this recipe uses the language 'caption' and 'subtitle' interchangeably. We'd prefer if this recipe just refers to captions, as subtitles should be identified separately from captions. Captions can still be made available as .srt files.

Comments from the AV Annotations TSG.

kirschbombe:
Since currently the cookbook is for v3 recipes, I might suggest creating a new recipe rather than updating any v3 recipes with the proposed `provides` property since it will not be valid in v3 manifests. We probably want to keep the v3 recipes intact?

PR #570, opened 2024-12-20
0219 - add proposed provides property and remove subtitles

elynema:
See issue #555 . First pass at adding the proposed  provides property to the captions recipe and removing references to subtitles. We propose to handle subtitles with a different value for the provides property in #556, so they need to be demonstrated in a different recipe.


glenrobson:
Direct link to the preview of the recipe:

https://preview.iiif.io/cookbook/update_caption_recipe/recipe/0219-using-caption-file/

glenrobson:
Changes look great. The only thing I noticed is you will have to change the title in the following file:

https://github.com/IIIF/cookbook-recipes/blob/update_caption_recipe/_includes/links.md

To update the main page at: https://preview.iiif.io/cookbook/update_caption_recipe/index.html

recipe/0219-using-caption-file/index.md
```diff
@@ -44,8 +44,9 @@ In this example we use a caption file in the WebVTT format, but other options in
 # Related recipes
 
 - [Simplest Manifest - Video][0003]
+- [Using Caption and Subtitle Files in Multiple Languages with Video Content][0074]
+- [A Side-by-side Transcript of a Video Recording][0253]
 - [Providing Access to Transcripts of A/V Content][0017]
-- [Providing Alternative Representations][0046]
 - [Using Annotations for Timed Text][0079]
```
saracarl, line 50: Any idea why this link isn't working on the preview site?

recipe/0219-using-caption-file/index.md
```diff
@@ -15,15 +15,15 @@ topic: AV
...
 
-In addition to this implementation, one should consider offering the captions or subtitles as multiple timed annotations, making the text available in multiple ways. See [Using Annotations for Timed Text][0079].
+In addition to this implementation, one can also offer captions as multiple timed annotations, making the text available in multiple ways. See [Using Annotations for Timed Text][0079].
 
 While captions, subtitles, and transcripts each present some text interpretation of the A/V content, the ways in which they are consumed by users differ. For a more detailed discussion about these differences see [Transcripts, Captions, and Subtitles - General Considerations][0231].
```
saracarl, line 28: Would we want to address `commenting` here as well?  Maybe link to that recipe?

saracarl (changes requested):
comments for consideration, figuring out the link is the most important bit.

recipe/0219-using-caption-file/index.md
elynema: @saracarl Which link are you referring to? I removed the link to the recipe for providing alternative representations because it didn't seem particularly relevant. I would argue that 0257 [Serving HLS Files] should be removed as well.

recipe/0219-using-caption-file/index.md
elynema: @saracarl In this particular section of the recipe, we're really only talking about captions, so I don't think `commenting` is relevant here. I could see potentially putting a link to a commenting recipe at the bottom?

recipe/0219-using-caption-file/index.md
saracarl: looks good -- maybe it just hadn't rebuilt?

recipe/0219-using-caption-file/index.md
saracarl: We decided to link to https://preview.iiif.io/cookbook/0103-poetry in the related recipes to "answer" this.

triplingual:
@glenrobson Does this PR need to get re-pointed to `v4` rather than the trunk?

glenrobson:
Yes, so I think we need to close this pull request and open a new one which takes the `update_caption_recipe` branch and points it to `v4`

glenrobson:
Created: https://github.com/IIIF/cookbook-recipes/pull/711

PR #572, opened 2025-01-02
Recipe #0561: Visible Text Annotation

### LINKED ISSUE #561: Adding visible text to a Canvas
## Recipe Name  
Adding visible text to a Canvas

## Use case  
You want to place visible textual content on a Canvas as a narrative element.

triplingual:
Closes #561


glenrobson:
Editors are happy for this to go through to the TRC but have the following comments:

# Implementation notes
 * The [Presentation API](https://iiif.io/api/presentation/3) states clearly that - no need to have link to prez spec
 * "Annotations referenced in items have to have the motivation value painting" - instead of "have to have" reword as must have 
 * We think that painting as a value should be in quotes rather than backticked.  

# Example 
 * Either in the use case or the example could you produce an image of what it should look like if implemented. This will help readers visulise the recipe and also for implementers to know what they are aiming for. 

# Related recipes
 * Remove: [Image Rotation Two Ways](https://preview.iiif.io/cookbook/0561-text-on-image/recipe/0040-image-rotation-service/) as the CSS uses annotation styling rather than HTML styling. 
 * Remove: [Image and Canvas with Differing Dimensions](https://preview.iiif.io/cookbook/0561-text-on-image/recipe/0004-canvas-size/) - Not sure this is related.

triplingual:
Just to note that I will start tackling this 20 March.

triplingual:
Highlighting that I changed the formatting of "painting" where used as an adjective but left as backticked when used as a motivation value.

triplingual:
Changes made per IIIF/trc/issues/133 comments

Issue #464, opened 2024-02-23
Re-use of Manifests


robcast:
## Recipe Name  
Re-use of Manifests

(Provide link to `index.md` of the issue, if available. To be filled in after issue is created - you need the issue number!)

## Use case  

I want to create a new Manifest based on parts of an existing manifest, either to mix images from different Manifests into one presentation or to add metadata or Annotations to an existing presentation. What rules should I follow, which elements should be copied into the new Manifest?

A motivation for this recipe was the discussion on Slack started by Kelli Babcock on 2024-02-09: https://iiif.slack.com/archives/C0AL6BU76/p1707488457989169

kirschbombe:
Would a use case for this be "I want to create a Range in my Manifest that uses Canvases from another Manifest"? I have that use case (although I don't think any viewers support that at the moment) - but it could be that I haven't constructed the range properly. I can share our test manifests for this for review and a starting point. I was planning to do this recipe at some point, but happy for someone else to do it instead :)

robcast:
Please share your example @kirschbombe. Constructing a Range could be an example use case. We should aim for a simple example to explain the general case as usual.

stephenwf:
I see in the recipe there is a human-readable way to see that the Manifest has been derived from another Manifest but should there be a machine-readable way to do the same thing? And could/should viewers show this more formally, similar to how GitHub handles "forks"

robcast:
@stephenwf unfortunately there is currently no machine readable solution in the spec to reference the source Manifest URI so we resorted to only using human-readable presentation metadata.

If enough people take up the issue that there should be a solution there could be a formal addition in the spec in the future ;-)

PR #588, opened 2025-03-07
Recipe on Manifest reuse.

robcast:
PR for recipe #464


glenrobson:
The cookbook editors are happy for this to go through to the TRC but request the following changes:

# Implementation notes
 * Change "You have to change" to "You must change" 
 * Change "You have to reproduce the requiredStatement" to something similar to you should reproduce the requiredStatement and depending on the licenses you may have to reproduce. 

# Example
 * Why not include more viewers? Shouldn't all of the viewers that support the [simple image](https://preview.iiif.io/cookbook/0464-reuse-manifest/recipe/0001-mvm-image/) be included?
 * Add code highlighting for the annotation because its new and the metadata because that points to the original.

PR #593, opened 2025-03-18
Update 0074 with provides

nfreire:
testing


glenrobson:
I think this is replaced by: https://github.com/IIIF/cookbook-recipes/pull/621

glenrobson:
We can re-open if we need to but I _think_ the changes are in the other pull request.

PR #596, opened 2025-03-18
0079 using annotations for captions

### LINKED ISSUE #79: Using Annotations for Timed Text

nfreire:


recipe/0002-mvm-audio/index.md
```diff
@@ -12,33 +12,31 @@ viewers:
  - Aviary
  - Theseus
  - Glycerine Viewer
+ - TIFY
+v4-viewers:  
+ - Mirador
+ - UV
+ - Aviary
```
stephenwf, line 19: I think Theseus now supports Audio Timelines now (and video?).

glenrobson:
Maybe we should point this to the v4 branch now rather than the master.

PR #604, opened 2025-05-21
0219 caption files with video content v3/v4

### LINKED ISSUE #555: Add provides property to Using Caption and Subtitle Files with Video Content recipe

elynema:
New version of 0219 to merge into v4 cookbook branch


glenrobson:
Comments from the editors:

## Use case
- [ ] “on top of your IIIF video resources.” is this giving too much direction to the viewer? Can we mention something about accessibility being the main driver rather then specify what the viewer should do?
- [x] Values of properties should be in speech marks rather than backticks, see "supplementing" 

## Restrictions

- [ ] @markpbaggett to check if the note on VTT for HLS is true 

## Example 
- [x] Manifests are switched. V3 shows v4 manifest v4 shows v3 manifest. 
- [x] Check validation Glen to check this as it doesn't seem to be a valid v4 manifest.

triplingual:
I'm holding off for the moment engaging more with the prescriptive language around what to do with the captions. With #617, I can believe there are many possible ways for a viewer to present a transcript file. With captions/subtitles, however, I'm no accessibility specialist but am inclined to agree with the current recipe text that "Captions are typically overlaid on top of the video." This feels like a question for accessibility experts, with advance apologies if the editors group includes such people and that's why the edit is proposed.

triplingual:
Change in 562ec52 addresses TRC comment, I think, @glenrobson

PR #617, opened 2025-06-17
0253 av transcript file v4

### LINKED ISSUE #253: AV Transcript which uses a supplementing annotation
## Recipe Name  
A side by side transcript of a video recording

## Use case  
I have a transcript that I would like to show along side the video. The transcript doesn't have timecodes but should be shown along side the recording.

elynema:
Pull request for new v4 version of recipe #253 Using Transcript Files with Audio or Video Content. There is no v3 version of this recipe.


triplingual:
Auto-close linkage should be dissolved if we decide to do both v3 and v4 versions of #253

glenrobson:
Comments from the editors:

## Title:
- [x] New title: Video with accompanying transcript 

## Use case:
- [x] Remove: side-by-side from use case as specifying what the view does with it is out of scope

## Implementation notes: 
- [x] Remove reference to vtt and srt files, as files with time codes are probably going to be [Using Caption Files with Video Content ](https://github.com/IIIF/cookbook-recipes/pull/604)
- [x] Remove docx and pdf references as its difficult to render these in a viewer. 
- [x] Introduce provides a bit more and say it doesn’t tell the viewer what to do with but once a viewer knows what the linked file is it can make an informed decision on how to display it.  
- [x] Reference the "Using Caption Files with Video Content" and say why its different 
- [x] Add note that linking an transcription using `supplementing`  was possible in v3 but it would have to be done without provides as this isn't supported in v3. 

## Example
- [x] Change to txt file rather than a docx as this seems to be an unnecessary barrier to implementation to have it as a binary file. 
- [x] Change label in the annotation from Transcript in WebVTT format to Transcript in txt format
- [x] I don't think the validation is working on this recipe. Glen to check 

OK to go through but editors would like to see it before it goes to the TRC.

Also note this discussion in the av channel:

https://iiif.slack.com/archives/C1HRK315L/p1783720692043259

recipe/0253-using-transcript-file/v4/manifest.json
```diff
@@ -0,0 +1,73 @@
...
+          "items": [
+            {
+              "id": "{{ id.path }}/canvas/page2/a1",
+              "type": "Annotation",
+              "motivation": ["supplementing"],
+              "provides": ["transcript"],
+              "body": {
+                "id": "https://fixtures.iiif.io/video/indiana/lunchroom_manners/lunchroom_manners.txt",
+                "type": "Text",
+                "format": "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
```
markpbaggett: The mime type should be updated to match the file type in the `id`.

markpbaggett:
@triplingual, apologies for responding so late before TRC next week.

I see you've already updated the annotation to be a plain text file.  I think that makes a lot of sense because even though some applications / viewers might be able to convert a DOCX it's much more likely that viewers will be able to universally display plain text.  The real benefit of this recipe, as compared to one having to do with timed text (SRT or VTT), is that it shows a viewer should display a transcript even if it isn't timed. This also reinforces that the modeling doesn't change based on whether the transcript is timed or not; whether it's a VTT, a DOCX, or now a plain text file, it's still an Annotation with `"provides": ["transcript"]`. The recipe is really demonstrating that this pattern holds regardless of format, while also being more easily implementable.

glenrobson:
Just in case anyone from the TRC looks a this issue. We went back and forth on whether we should use a docx or a text file. We decided to leave it as a text file as it should be easier to implement. 

We did see use cases for docx transcriptions and Ramp has implemented this in this example:

https://media.dlib.indiana.edu/media_objects/5x21tm82q

Using a JS library called [mammoth](https://www.npmjs.com/package/mammoth) which converts a docx file to HTML. See [transcript-parser.js](https://github.com/samvera-labs/ramp/blob/main/src/services/transcript-parser.js#L444)

recipe/0253-using-transcript-file/v4/manifest.json
triplingual: Yes -- an oversight, and thanks for noticing.

triplingual:
Last three commits address TRC comments, I believe, @glenrobson.

PR #621, opened 2025-07-22
V4 0074 update with provides

glenrobson:


triplingual:
(Commit message for 3f5fae5 should have a paintbrush emoji rather than ???)

PR #625, opened 2025-07-29
0045 CSS in Annotations

### LINKED ISSUE #45: CSS styling
## CSS styling
(this will usually be the same as the issue name)

## Use case  
..

triplingual:
Closes #45


glenrobson:
Comments from the cookbook authors:

The title is CSS in an Annotation Body but now we have css in the body and target this should be renamed. (Glen suggestion: Styling Annotations with CSS? maybe)

## Use case

Could you do the use case from with the use of 'you' e.g:

You have two different authors… You would like to distinguish the authors. 

To match the style of other use cases. 

## Implementation notes

"Using CSS to alter resource presentation styles is not specified in the [ IIIF Presentation 3.0 API](https://iiif.io/api/presentation/3.0/) except insofar as the Presentation API incorporates types from the W3C Web Annotation Data Model."

Could you reword as something like:

Using CSS in annotations is covered in the W3C Web Annotation Data Model 

This is due to some annotation stuff being covered in the spec but not CSS. 

With the second paragraph on  inline, with an inline stylesheet, or with an external stylesheets the editors found this confusing and wondered if it could be re-written to focus first on the method you are using in the example (external stylesheet) and then at the end briefly mention the other options and link to the recipes. Possibly by moving the "see [Image Rotation Two Ways](https://preview.iiif.io/cookbook/0045-css/recipe/0040-image-rotation-service/) (inline stylesheet) and [Visible Text Resource on a Canvas](https://preview.iiif.io/cookbook/0045-css/recipe/0561-text-on-image/) (inline CSS)." from the example up to this paragraph.  

## Example
 * Remove Python: iiif-prezi3 
 * Although thesesus doesn’t show the viewer it would be good to include it as partial support with a note in the example mentioning the target isn't working.

OK to go to the TRC with changes.

recipe/0045-css/index.md
```diff
@@ -0,0 +1,57 @@
+---
+title: CSS in an Annotation
+id: 45
+layout: recipe
+tags: [style]
+summary: "Using an external CSS stylesheet in an annotation body, annotations can be styled in limited ways"
+viewers:
+ - Theseus
```
glenrobson, line 8: I think this should be:

 - id: Theseus
   support: partial
   
As it doesn't support the target styling.

recipe/0045-css/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+One example: Since IIIF Canvas dimensions are unit-less, using pixels for text size is valid but may be interpreted variably across viewers or other clients.
+
+## Example
+
+This recipe focuses on annotations with motivations other than painting and on an external CSS stylesheet. The Theseus viewer is included here for its support of styling the annotation text, but it does not currently support styling the `target`.
+
+{% include manifest_links.html manifest="manifest.json" %}
+
+{% include jsonviewer.html src="manifest.json" config='data-line="56,60,71,78,82,93"' %}
```
glenrobson, line 43: I think this should be:

"56,60,72,79,83,95"

As the highlighting is slightly off

recipe/0045-css/index.md
stephenwf: The target styling is supported now on Theseus, but there is a CORS issue when loading the `style.css` file at the moment. https://theseusviewer.org/?iiif-content=https%3A%2F%2Fpreview.iiif.io%2Fcookbook%2F0045-css%2Frecipe%2F0045-css%2Fmanifest.json

PR #638, opened 2025-09-15
0027 alternative page order

glenrobson:


glenrobson:
## Use case: 

Change:
current form to current order 

## Implementation Notes

Add a paragraph explaining the following from the spec:

"user interfaces that interact with this order should use the order within the selected Range, rather than the default order of items."

from the sequence text in https://iiif.io/api/presentation/3.0/#behavior

Add a paragraph explaining that behaviour of a range can be a list and could include individuals or paged alongside sequence if you would like the range canvases to be shown as paginated. Also note ranges can have `viewingDirection` in case of right to left ordering of pages. 

## Restrictions: 

Sequence ranges need to be top order ranges (can’t be embedded in a table of contents). Text from specification:  

“Ranges that have the behavior value sequence must be directly within the structures property of the Manifest, and must not be [embedded](https://iiif.io/api/presentation/3.0/#12-terminology) or [referenced](https://iiif.io/api/presentation/3.0/#12-terminology) within other Ranges." from https://iiif.io/api/presentation/3.0/#54-range 

## Example

 * Example: Context should be a string 
 * Use minimum amount of pages to show issue. (~6)
 * The editors would prefer real manuscript images. As the recipe is showing out of order as sequence a and in order as sequence b you could potentially use the picart_german images from UCLA: https://fixtures.iiif.io/. The canvas labels should show the page numbers to demonstrate the order.  

## Related recipes
 
Add a note to the table of contents recipe to say its an example of different way of using ranges. (for this type of note see: https://preview.iiif.io/cookbook/0045-css/recipe/0045-css/)

OK to go to TRC if its possible to get the items above done.

leontpng:
@glenrobson @giacomomarchioro I merged the post-TRC edits to the 0027 branch.

PR #641, opened 2025-10-07
Add TIFY viewer

t11r:
This adds TIFY to each supported recipe of the IIIF Cookbook.

Website: https://tify.rocks/
GitHub repository: https://github.com/tify-iiif-viewer/tify


t11r:
Thank you for reviewing, @glenrobson! I added TIFY to the second example of 0010 and removed it from 0014 and 0035. I will add those with a separate PR once the issues are fixed.

PR #651, opened 2025-11-21
Adjust site text for new Mirador version

triplingual:
Since we have been shifted to using v4, we should remove specifying v3 in website text. I think we should not mention the Mirador version at all, in the same way we don't for most other viewers. Whether to retain it for UV is up for discussion.


triplingual:
(Turns out there's not much text that uses the Mirador version number. There are other changes to recipe text in #590, and in some world this change should just be with those, but in this world they are in 2 different PRs.)

glenrobson:
Merging as the build issue is unlreated to the change.

PR #670, opened 2026-02-26
Synchronize recipes with iiif-prezi3 support.

### LINKED ISSUE #671: Synchronize IIIF Prezi3 support with recipes.
As we're trying to add support for the entire cookbook in iiif-prezi3, I thought it would be a good idea to go ahead and update the cookbook to point at current solutions.  You can find these in the pull request below:

https://github.com/IIIF/cookbook-recipes/pull/670

markpbaggett:
# What Does This Do?

This adds iiif-prezi 3 links to all recipes that are currently supported and removes one where it's not supported currently.


glenrobson:
Closing in favour of https://github.com/IIIF/cookbook-recipes/pull/693 which has the actions running.

PR #680, opened 2026-04-17
0025 Newspaper Article Index

### LINKED ISSUE #25: Table of contents (ranges) - articles in a newspaper
## Table of contents (ranges) - articles in a newspaper

## Use case  
Logical structure of newspaper down to article level

leontpng:
First recipe draft


stephenwf:
This is working in Theseus from the range panel (if you click on a range)
https://theseusviewer.org/?iiif-content=https%3A%2F%2Fraw.githubusercontent.com%2FIIIF%2Fcookbook-recipes%2Fa3841edebce4db8b6d3a19d57b7be211a2c0cb95%2Frecipe%2F0025-newspaper-article-index%2Fmanifest.json

The viewer control
image

The highlight view
image

There is also a sequential view where you can read the article columns in a single column scrolling down.

cool recipe!

glenrobson:
Recipe link: https://preview.iiif.io/cookbook/0025-newspaper-article-index/recipe/0025-newspaper-article-index/

recipe/0025-newspaper-article-index/index.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Newspaper Article Index
+id: 25
+layout: recipe
+tags: [text, image, newspaper]
+summary: "Using Ranges to build a navigable index for newspaper articles spanning multiple non-consecutive pages"
+viewers:
```
stephenwf, line 7: Can Theseus be added to the supported list?

glenrobson:
The cookbook editors are happy for this to go to the TRC but have requested the changes below. One of the big changes they've requested is to remove the link to the annotations with the supplementary field in the range. They think the addition of supplementary would be better in a separate recipe. This also means Theseus can be added as a viewer that supports this recipe. 

## Title

Could the recipe be remanned from "Newspaper Article Index" to "Navigation by Newspaper Article" 

## Use case

Rename "Use Cases" as "Use Case" to match other recipes.  

## Implementation notes

Could you add the following to the end of paragraph 1:
These ranges inform the viewer that the resources are intended to be consumed in a particular order. 

2nd paragraph says:

"every entry in the items property of a Range must be a JSON object and have both id and type properties"

but this isn't true as SpecificResource objects in the example have no id. Maybe remove this sentence? I can see that the text comes from the spec but its in the context of top level collections and might not be true for sub ranges which this one is. 

Keep the paragraph on "supplementary content such as OCR transcriptions" even though we are removing it from the example. 

Could you redirect the link from the spec to [https://iiif.io/api/cookbook/recipe/0309-annotation-collection/](https://iiif.io/api/cookbook/recipe/0309-annotation-collection/) in the following sentence:

"More on Annotation Collections [here](https://iiif.io/api/presentation/3.0/#58-annotation-collection)." 

Also its not good practice to use the word `here` as a link for accessibility reasons. So perhaps it should be:

"For more information on annotation collections see the [Grouping Annotations into Collections] recipe. 

## Example

Could you mention there is a second article in the newspaper (Das Turnier) which is on a single page. 

Could you remove the sentence:

"There is no predetermined display behavior for partial Canvases: the viewer may crop the Canvas or highlight the region of interest to direct the user." 

as the spec doesn't give guidance on viewer behaviour and this is true of all recipes. 

Remove “Each article Range is linked to an Annotation Collection that contains the OCR transcription of each article.” now that we've removed the annotation collections. 

## Manifest 

Update highlighting to include structures (lines 324-443)

Remove the two supplementary fields:

```
"supplementary": {
    "id": "https://preview.iiif.io/cookbook/0025-newspaper-article-index/recipe/0025-newspaper-article-index/annotations/5ee30fe6-cc3d-431a-9acf-2a715b770306.json",
    "type": "AnnotationCollection"
}
```

Issue #309, opened 2022-04-08
Annotation Collections


robcast:
## Recipe Name  
Annotation Collections

PR: #521 

index.md: https://preview.iiif.io/cookbook/309-annotation-collections/recipe/0309-annotation-collection/

## Use case  
You have a large number of annotations for multiple Canvases in a Manifest that should be displayed in a viewer optionally as a "layer".

robcast:
the W3c Web Annotation spec describes AnnotationCollections in https://www.w3.org/TR/annotation-model/#annotation-collection

The IIIF Prezi spec describes AnnotationCollections in https://iiif.io/api/presentation/3.0/#58-annotation-collection 

The Prezi spec only allows an AnnotationCollection in the `supplementary` property of a Range https://iiif.io/api/presentation/3.0/#supplementary

robcast:
Use-case: I have a web-service that identifies illustrations in scans early-modern printed books and provides a classification and the coordinates of these illustrations on the page. The service takes a IIIF Manifest as an input and provides Web Annotations targeting the Canvases from the Manifest as an output.
 
I want to save these Annotations in an AnnotationCollection and provide these in a viewer as an optional "layer" on top of the normal presentation of the Manifest. 

The user should see that this layer is available, she should see some information about the layer, and she should be able to show or hide the layer. The Annotations in the layer should be differentiated from other Annotations by their color or other means.

robcast:
Notes:
- references to Canvas-IDs (if the AnnotationCollection is not part of the Manifest) need to have a `partOf` providing a Manifest for the viewer to resolve the Canvas (or the Canvas-URI resolves itself)

robcast:
related issues:
- #306 about referencing the Manifest from the Canvas URI in the annotation
- UI ideas in Mirador https://github.com/ProjectMirador/mirador/issues/2946
- question about classifying annotations https://github.com/ProjectMirador/mirador/issues/3100

kirschbombe:
> Notes:
> 
> * references to Canvas-IDs (if the AnnotationCollection is not part of the Manifest) need to have a `partOf` providing a Manifest for the viewer to resolve the Canvas (or the Canvas-URI resolves itself)

Hi, @robcast -- As it is currently defined by the spec, AnnotationCollections are not part of a Manifest, they are only referenced.

robcast:
I saw the verdict in https://github.com/IIIF/api/issues/2203. I'll think about it and try to come up with an example for the recipe.

Issue #620, opened 2025-07-18
309: Annotation collection - can we add a second annotation collection?


glenrobson:
## Recipe info:

 * Name: Grouping Annotations into Collections
 * link: https://iiif.io/api/cookbook/recipe/0309-annotation-collection/

## Correction request  
Currently there is only one collection which makes testing if it works in a viewer a bit difficult. If it had two annotation collections then you could tell if a view can switch between them.

Second annotation colleciton could be transcription which we have from the newspaper recipe.

PR #681, opened 2026-04-24
Added Theseus to supported viewer for annotation collections (#309)

stephenwf:
# Theseus support for annotation collections

Cookbook link: https://iiif.io/api/cookbook/recipe/0309-annotation-collection/

image

If there are one or more annotation collections, they will be visible in the annotation sidebar, where you can give each different colours. (Also available on the canvas itself under the annotations menu).

You can also toggle on/off different collections.

https://theseusviewer.org/?iiif-content=https%3A%2F%2Fiiif.io%2Fapi%2Fcookbook%2Frecipe%2F0309-annotation-collection%2Fmanifest.json&panel=annotations

### DIFF
--- a/recipe/0309-annotation-collection/index.md
+++ b/recipe/0309-annotation-collection/index.md
@@ -6,6 +6,7 @@ tags: [tbc]
 summary: "tbc"
 viewers:
  - Glycerine Viewer
+ - Theseus
 topic: 
  - basic
 ---


glenrobson:
Note this issue: https://github.com/IIIF/cookbook-recipes/issues/620 

Also discussed what happens if an annotationpage is part of two collecitons...

Issue #540, opened 2024-09-06
Share a link for opening two or more Canvases each from a different Manifest


giacomomarchioro:
## Recipe Name  
Share a link for opening two or more Canvases, each from a different Manifest

## Use case  
I want to compare pages from two different manuscripts and share a link to open both of them on the same view. I want to share a link to a colleague showing two similar paintings in two different collections. I want to save my current workspace (open Canvases in the current viewer) for later users. I want to open my current workspace with another viewer. 

Mirador already implements a format for exporting the current workspace for sharing or later users. Content State API could be used for the same purpose, adding the advantage of direct loading of the workspace using a crafted link with the `iif-content` query parameter. The [multiple targets for a comparison view section](https://iiif.io/api/content-state/1.0/#53-multiple-targets-for-a-comparison-view)) describes a method for targetting two Canvas at the same time; each Canvas could be from a different Manifest.

## Other considerations

We can also assume that viewers that implement [the Content State API section regarding the targeting of a Canvas region](https://iiif.io/api/content-state/1.0/#51-a-region-of-a-canvas-in-a-manifest) will be able to open the Canvas and also draw attention to the selected region. 

We assume that the two Canvas will be displayed side by side, but could viewers concatenate the two Manifests or open two separate instances of the viewers?

Should we concerned about the length of the final base64URL encoded URL? (it seemed Apache and Internet Explorer were the bottlenecks, with approx. 4000 and 2000 characters, while other technologies usually support up to 40000).

giacomomarchioro:
I was thinking to use these two images for creating two different manifests to be loaded at the same time: 

https://fixtures.iiif.io/info.html?file=/images/Yale/ycba/The_Colosseum_Rome.jpg

https://fixtures.iiif.io/info.html?file=/images/Yale/ycba/The_Colosseum_Rome2.jpg

Any other suggestion is welcomed.

giacomomarchioro:
As discussed in the cookbook meeting I'll use this manifest from another recipe :

https://preview.iiif.io/cookbook/0485-contentstate-canvas-region/recipe/0318-navPlace-navDate/manifest-2.json

and I'll create another one showing another view of the Colosseum that will be store with the Content State example.

PR #686, opened 2026-04-24
Added Theseus support for opening multiple canvases (#540)

stephenwf:
Cookbook link: https://iiif.io/api/cookbook/recipe/0540-link-for-opening-multiple-canvases/

This is a strange one. Because Theseus doesn't have a comparison view, I've chosen instead to create a "virtual" manifest with only resources listed. So for example, if there are 2 canvases from 2 different manfiests, it will show you only those two canvases (with required statements/rights hoisted up to the canvas). However, if you load a few canvases from the same manifest, it treats it like a sort of "filtered" view.

I think there are other cases to consider, and possibly new recipes. I'm happy with the virtual manifest for Theseus. This could be used to share "annotations I worked on" from an annotation platform, or similar - or as a form of bookmarking.

Here is the content state in the recipe:
https://theseusviewer.org/?iiif-content=https%3A%2F%2Fiiif.io%2Fapi%2Fcookbook%2Frecipe%2F0540-link-for-opening-multiple-canvases%2Fannotation.json

(Works if you paste in the encoded content state too)

image

### DIFF
--- a/recipe/0540-link-for-opening-multiple-canvases/index.md
+++ b/recipe/0540-link-for-opening-multiple-canvases/index.md
@@ -5,6 +5,7 @@ layout: recipe
 tags: [annotation, content-state]
 summary: "Allows users to use Content State API to open two canvases at the same time."
 viewers:
+  - Theseus
 topic:
  - content-state
 ---


glenrobson:
Need to move locally to get preview to check the viewer links work correctly.

We think this is good to merge but as a user experience is there a way to access the two manifests? In this example they are only single images but if you imagine they were two books with a page selected on each, how would you navigate to the other pages for each book? 

Do you know if this would work with AV resources?

stephenwf:
Yes, maybe a way to get back to the source Manifest makes sense (perhaps in a new window?). It should work for AV, it doesn't currently allow for cropping the AV resource though - maybe that could be a future recipe? 

For the "link back to the source Manifest" - I think that could be supported in a IIIF way (since a virtual Manifest is created). That would be a useful feature for the Manifest Editor too - keeping a link back to the source manifest in a detectable way so it can be surfaced. I feel like I have seen a recipe for that somewhere

glenrobson:
Closing in favour of: https://github.com/IIIF/cookbook-recipes/pull/686

Issue #599, opened 2025-04-16
Drag and Drop


glenrobson:
## Recipe Name  
It would be useful to have a recipe that describes the Content state drag an drop pattern that includes the IIIF logo which is correctly configured so that it can be opened by a viewer.

It will also be useful to show which viewers support drag and drop in the viewer matrix. 

## Use case  
A user may want to take a manifest from an instution and open it into a IIIF viewer which maybe configured to help with their research.

PR #687, opened 2026-04-24
Added Theseus support to drag and drop (#599)

stephenwf:
Not really a link to follow, but you can drag the example from the cookbook:
https://iiif.io/api/cookbook/recipe/0599-drag-and-drop/

On to any page on Theseus to load it (incl. homepage):
https://theseusviewer.org/

I've chosen _not_ to b64 encode the content state in the URL that is generated, but maybe thats a conversation to be had.

> !NOTE
> You can also generate content state and drag and drop links from Theseus share menu. Its a bit rough right now, but it seems to work.

image

(I'll get the logo added!)

### DIFF
--- a/recipe/0599-drag-and-drop/index.md
+++ b/recipe/0599-drag-and-drop/index.md
@@ -5,6 +5,7 @@ layout: recipe
 tags: [content-state]
 summary: "A model for manifest publishers and viewer developers for how to facilitate and accommodate dragging a IIIF manifest from one application to another."
 viewers:
+  - Theseus
 topic:
  - content-state
 ---


recipe/0599-drag-and-drop/index.md
glenrobson, line 83: Need to change this line and add a link to a viewer.

glenrobson: Maybe try:

```
{% include manifest_links.html " %}
```

with no viewers it should take the viewers from the front matter.

Not sure what will happen if we don't supply a manifest.

glenrobson (changes requested):

recipe/0599-drag-and-drop/index.md
triplingual: This is a real pickle. Having no `manifest` attribute is the same as having `manifest=""`, which just sends you to the viewer with a malformed `iiif-content` parameter. Also, it may turn out that the way to see the recipe/functionality in action is slightly different from viewer to viewer. Perhaps replace that line with this instead?
```
### Viewer Support

+ Theseus: Open [the main Theseus page](https://theseusviewer.org/) in a new browser window or portion of a split view, drag the IIIF logo below over that new space, and drop it anywhere in the viewport.
```

triplingual, line 110: Can delete these as part of this PR, yes?

triplingual (changes requested):

recipe/0599-drag-and-drop/index.md
glenrobson: Group agree to just hard code it as this is the only recipe without either content state or a Manifest.

glenrobson: Yes I think we can get rid of this now.

glenrobson:
Closing in favour of https://github.com/IIIF/cookbook-recipes/pull/706

PR #689, opened 2026-04-28
688-Stereograph-Images

### LINKED ISSUE #688: Stereogram Images Presentation
## Use case
  
You have an image of a stereogram, such as a card originally intended for a physical viewer that splits the visible portion between the user's eyes to simulate a locked 3D perspective. Rather than just presenting the full image for inspection, you wish to define a view which flickers between each side to suggest the depth recorded.

There are many ways to consume this visual apparition, but the oscillating presentation is selected as possible within IIIF and one that can be contained to the viewer, rather than relying on the user to have VR or anaglyph equipment or to cross their eyes just so for "magic eye" style stereographic viewing.

- **[Stereogranimator](https://stereo.nypl.org/)** is an archived presentation of many of these types of resources in different ways.
- **[IIIF Stereographs](https://stereograph.davidnewbury.com/)** reproduces several of these visualizations with IIIF Image resources, but without using IIIF Presentation.
- **[Fort Marion (stereograms)](https://collections.library.yale.edu/catalog/2018526)** is a collection of IIIF Resources that may benefit from this type of recipe.

cubap:
Create a Yale resource Manifest that splits the image and shows it in the intended way as an alternating stereogram.


recipe/0688-stereogram-timeline/manifest.json
```diff
@@ -0,0 +1,87 @@
...
+                "type": "Image",
+                "format": "image/jpeg",
+                "height": 1628,
+                "width": 1409,
+                "service": [
+                  {
+                    "@id": "https://collections.library.yale.edu/iiif/2/1073279",
+                    "@type": "ImageService2",
+                    "profile": "http://iiif.io/api/image/2/level2.json"
+                  }
```
glenrobson: Are you planning to use a image API selector for the region of the stereogram?

mathewjordan: I was thinking the same @glenrobson. There are two methods.

1) Remove the Image service and lean on xywh in the `id` of the body.
2) Retain the Image service and require support of https://iiif.io/api/cookbook/recipe/0299-region/

cubap: This was an odd one. The viewers don't really like selectors on the Canvas, even though it is valid. I ultimately had to swap it out with the Image API alone to get it to [run in Clover](https://samvera-labs.github.io/clover-iiif/docs/viewer/demo?iiif-content=https%3A%2F%2Fraw.githubusercontent.com%2FIIIF%2Fcookbook-recipes%2Frefs%2Fheads%2F0688-sterogram-images%2Frecipe%2F0688-stereogram-timeline%2Fmanifest-umedia-856.json). After experimenting ImageAPISelector on the Canvas and direct images were the only ways that worked in a viewer. However, using the direct images worked minimally everywhere and the ImageApiSelector was very poorly supported (as seen in the Viewer Matrix)
image

Should I include them both with different lists of viewers? Should we prefer the more broken one as more "pure"? Cards on the table, Clover is the only one it is completely working as expected.

mathewjordan: @cubap, I like the dual examples at https://preview.iiif.io/cookbook/0688-sterogram-images/recipe/0688-stereogram-timeline/. Switching to the fish example seems like a good call as well. These seem to work better when there is an item in the foreground. 

One suggestion, I feel like it could add context if we add an image of the original stereograph in the markdown content of the recipe.

Original stereograph

cubap:
I wonder how compliant the viewers need to be to be included? Something like [Mirador](https://projectmirador.org/demo/?iiif-content=https%3A%2F%2Fdevstore.rerum.io%2Fv1%2Fid%2F6a107b97ceaf4320a17f2c64) load and "play" automatically, but won't loop or reset, so you never get a good "flicker" experience. 

The compliance between the ImageApiSelector and the Image API regions is different and I would like to call it out, but this is a messy corner of use cases.

recipe/0688-stereogram-timeline/manifest.json
cubap: Thanks for the feedback. I have updated it to incorporate the suggestions.

mathewjordan:
> I wonder how compliant the viewers need to be to be included? Something like [Mirador](https://projectmirador.org/demo/?iiif-content=https%3A%2F%2Fdevstore.rerum.io%2Fv1%2Fid%2F6a107b97ceaf4320a17f2c64) load and "play" automatically, but won't loop or reset, so you never get a good "flicker" experience.
> 
> The compliance between the ImageApiSelector and the Image API regions is different and I would like to call it out, but this is a messy corner of use cases.

Mirador supports multiple painting annotations on a single Canvas so the _load and "play"_ is the second Annotation loading over top of the first one. The _play_ in this case is due to user's network request handling of each asset and doesn't appear much at all at higher bandwidth. 

It is not respecting `duration` or the `behavior` value of `repeat` so my two cents would be a no. This would be similar to support of https://iiif.io/api/cookbook/recipe/0560-resources-on-a-timeline/ optionally compounding with https://iiif.io/api/cookbook/recipe/0299-region/.

mathewjordan:
One more thought... One requirement may need to be that a user should be able to: 1) toggle between all frames 2) pan and zoom each of the images.

PR #699, opened 2026-05-08
Add newspaper article

glenrobson:
First stab at adding the newspaper articles

### DIFF
--- a/recipe/0025-newspaper-article-index/manifest.json
+++ b/recipe/0025-newspaper-article-index/manifest.json
@@ -341,16 +341,37 @@
           },
           "items": [
             {
-                "id": "{{ id.path }}/canvas/p2#xywh=553,1157,470,1103",
+              "type": "SpecificResource",
+              "source": {
+                "id": "{{ id.path }}/canvas/p2",
                 "type": "Canvas"
+              },
+              "selector": {
+                "type": "FragmentSelector",
+                "value": "xywh=553,1157,470,1103"
+              }
             },
             {
-              "id": "{{ id.path }}/canvas/p2#xywh=1018,95,461,245",
-              "type": "Canvas"
+              "type": "SpecificResource",
+              "source": {
+                "id": "{{ id.path }}/canvas/p2",
+                "type": "Canvas"
+              },
+              "selector": {
+                "type": "FragmentSelector",
+                "value": "xywh=1018,95,461,245"
+              }
             },
             {
-              "id": "{{ id.path }}/canvas/p5#xywh=580,105,460,2165",
-              "type": "Canvas"
+              "type": "SpecificResource",
+              "source": {
+                "id": "{{ id.path }}/canvas/p5",
+                "type": "Canvas"
+              },
+              "selector": {
+                "type": "FragmentSelector",
+                "value": "xywh=580,105,460,2165"
+              }
             },
             {
               "type": "SpecificResource",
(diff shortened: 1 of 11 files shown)


recipe/0025-newspaper-article-index/manifest.json
```diff
@@ -323,91 +323,32 @@
...
-              "selector": {
-                "type": "FragmentSelector",
-                "value": "xywh=1069,119,506,2109"
-              }
-            }
-          ]
+            "{{ id.path }}/canvas/p2#xywh=553,1157,470,1103",
+            "{{ id.path }}/canvas/p2#xywh=1018,95,461,245",
+            "{{ id.path }}/canvas/p5#xywh=580,105,460,2165",
+            "{{ id.path }}/canvas/p5#xywh=1044,104,466,2147"
```
glenrobson: Change one of these to be the full specific resource example

glenrobson:
Recipe: https://preview.iiif.io/cookbook/add_newspaper_article/recipe/0025-newspaper-article-index/

recipe/0025-newspaper-article-index/manifest.json
glenrobson: Turns out they can't be simple strings. They at least must be objects containing links to canvases. e.g:

```
{
     "id": "http://localhost:4000/recipe/0025-newspaper-article-index/canvas/p2#xywh=553,1157,470,1103",
     "type": "Canvas"
},
```

The spec mentions:

"Parts of Canvases may also be identified by appending a fragment to the Canvas’s URI, and these parts are still considered to be Canvases: their type value is the string Canvas"

From https://iiif.io/api/presentation/3.0/#53-canvas but just before this it says:

"Parts of Canvases may be described using a Specific Resource with a Selector, following the patterns defined in the [Web Annotation](http://w3.org/TR/annotation-model/) data model. The use of the FragmentSelector class is recommended by that specification, as it allows for refinement by other Selectors and for consistency with use cases that cannot be represented using a URI fragment directly. Parts of Canvases can be [referenced](https://iiif.io/api/presentation/3.0/#12-terminology) from Ranges, as the body or target of Annotations, or in the start property."

So maybe we should use the SpecificResource and FragmentSelector option.

recipe/0025-newspaper-article-index/manifest.json
```diff
@@ -323,91 +323,51 @@
...
-                "value": "xywh=1069,119,506,2109"
+                "value": "xywh=1044,104,466,2147"
               }
             }
-          ]
+          ],
+          "supplementary": {
+            "id": "{{ id.path }}/annotations/5ee30fe6-cc3d-431a-9acf-2a715b770306.json",
+            "type": "AnnotationCollection"
+          }
```
stephenwf, line 391:
```suggestion
          "supplementary": [{
            "id": "{{ id.path }}/annotations/5ee30fe6-cc3d-431a-9acf-2a715b770306.json",
            "type": "AnnotationCollection"
          }]
```

glenrobson: I think it has to be an object:

"The value must be a JSON object, which must have the id and type properties, and the type must be AnnotationCollection."

From https://iiif.io/api/presentation/3.0/#supplementary

glenrobson:
@leontpng I think this is good to go. If it looks OK to you I will merge it into your branch.

PR #678, opened 2026-04-16
Update viewer support.md

leontpng:
#668 Theseus now supports alternative page sequences!

Issue #668, opened 2026-02-13
Add Theseus to the compatible viewers in recipe 0027-alternative-page-order


giacomomarchioro:
Theseus now supports alternative page orders using Range and `sequence` as described in 0027-alternative-page-order:

https://theseusviewer.org/?iiif-content=https%3A%2F%2Fiiif.io%2Fapi%2Fcookbook%2Frecipe%2F0027-alternative-page-order%2Fmanifest.json

We should add it to the recipe:

https://github.com/IIIF/cookbook-recipes/blob/c06c83b0126a05a4fcb3068a2e36c0dd56a7128d/recipe/0027-alternative-page-order/index.md?plain=1#L7

https://github.com/IIIF/cookbook-recipes/blob/c06c83b0126a05a4fcb3068a2e36c0dd56a7128d/recipe/0027-alternative-page-order/index.md?plain=1#L32

glenrobson:
Done in: https://github.com/IIIF/cookbook-recipes/pull/678

PR #701, opened 2026-05-08
0027 theseus

glenrobson:
#668 Theseus now supports alternative page sequences!

From https://github.com/IIIF/cookbook-recipes/pull/678


glenrobson:
Recipe URL: https://preview.iiif.io/cookbook/0027-theseus/recipe/0027-alternative-page-order/

Issue #702, opened 2026-05-08
Look at image selector rotation angle


glenrobson:
## Recipe info:

 * Name: Image Rotation Two Ways
 * link: https://iiif.io/api/cookbook/recipe/0040-image-rotation-service/

## Correction request  

Should it be 270 or 90?

glenrobson:
Fixed in: https://github.com/IIIF/cookbook-recipes/pull/703

PR #703, opened 2026-05-11
0040 Fix rotation angle in service manifest

triplingual:
Unclear how none of us caught this. IIIF Image API v3 clearly states that image rotation is clockwise. 270º clockwise (the previous value in this recipe) would leave the image upside down.

There may yet be an issue with bounding box and annotations the cookbook will have to address


triplingual:
Closes #702

PR #709, opened 2026-06-15
[0318-navPlace-navDate] Update last navDate

ksclarke:
Looks like there was a recent-ish [commit](https://github.com/IIIF/cookbook-recipes/commit/2041685618372c5afc335d47362d500d33a72cc8) to update the `navDate`(s) in this recipe, but there was one that was missed. This PR fixes the last straggler.


glenrobson:
Thanks for the pull request. I've merged it.

PR #711, opened 2026-06-15
0219 Update caption recipe

glenrobson:
From https://github.com/IIIF/cookbook-recipes/pull/570

See issue #555 . First pass at adding the proposed  provides property to the captions recipe and removing references to subtitles. We propose to handle subtitles with a different value for the provides property in #556, so they need to be demonstrated in a different recipe.


glenrobson:
I'm closing this pull request as I believe its been superseded by this one:

https://github.com/IIIF/cookbook-recipes/pull/711

which contains updates to the v3 recipe and also a new v4 recipe.

PR #721, opened 2026-09-15
Update viewer matrix support in light of Triiiceratops 1.0 release

d-flood:
- Triiiceratops now has support for audio and video manifests, which represents the bulk of significant changes
- Updated viewer base URL from the default GitHub pages URL to `https://triiiceratops.org/viewer/`. The default GitHub pages URLs continue to work, however.
- I think Triiiceratops was accidentally removed from the basic image recipe during the v4 branch merge, so I've restored that also.


_includes/viewer_link.html
```diff
@@ -73,7 +73,7 @@
     {% assign default_text="View in TIFY" %}
 {% elsif include.type == 'Triiiceratops' %}
     {% capture viewer_url %}
-    https://d-flood.github.io/triiiceratops/viewer/?iiif-content={{manifest_url |strip}}
+    https://triiiceratops.org/viewer/?iiif-content={{manifest_url |strip}}
```
d-flood, line 76: I'll note that both these URLs work because GitHub pages redirects to a custom URL. So, there isn't urgency from that perspective.

recipe/0229-behavior-ranges/index.md
d-flood, line 1: I interpreted the `noNav` to mean that a table of contents structural panel shouldn't be used to display the thumbnails. I see that Theseus does keep the ToC panel, though, so let me know if I've misinterpreted the intention.

recipe/0599-drag-and-drop/index.md
d-flood, line 1: After opening this PR I saw that there is one to add Theseus to this recipe also. I don't think that PR updates the body of the page, though.

I've set this up so that the specified viewers get bare viewer links, e.g., no `?iiif-content=` query param and to remove the "No viewers currently support this..." text. Happy to revise if a different approach is added to the Theseus drag and drop PR.

recipe/0229-behavior-ranges/index.md
stephenwf: I think the `no-nav` is only on the first 00:00-00:09 sub-range, which is excluded from the ToC list. It was a little ambiguous if `thumbnail-nav` should exclude it from the ToC though, I left it in just as an option. If another recipe comes along with range summaries or something then it could be useful to see both

recipe/0599-drag-and-drop/index.md
stephenwf: Yeah #706 has been hanging for a while. Now there are 2 implementations that support it, it might be clearer how to structure the page.

d-flood: I've just updated my branch to also include Theseus in the markdown frontmatter so now the page renders like this:

image

